PR #2, opened 2020-04-06
New Images

blairdrummond:
Take a look at these - I indicated the sizes of them. We can keep working on shrinking them down or taking subsets of features. 

# R image
Basically the R image is a smallish (8gb) image with geospatial stuff. It would be better if we could use

https://github.com/rocker-org/geospatial

But I wasn't successful in getting it to play nice with kubeflow :-/ 

The image also has minimal python+julia installs.

# all-dressed

A real kitchen sink. I would be ok with splitting it into a couple of different images. It has

Tensorflow, Pytorch, Keras, XGBoost, and the geospatial libraries and R.

A good option might be to ditch R and just use python, but it seems like R is actually pretty small. It's the ML libraries and the geomatics stuff that is so big. This image is about 15gb

# No-geo

This is just the all-dressed image, but with the geomatics stuff removed. It's about 12gb


blairdrummond:
In terms of making the images smaller, I think that we could probably think about splitting up the Python ML libraries (Don't *really* need both TF and PyTorch on the same image), and adding better cache cleanups after the pip and conda installs. I think those were the real culprit.

PR #4, opened 2020-04-07
Move geomatics stuff out of minimal into a geomatics --- rename r

blairdrummond:
Move geomatics stuff out of minimal into its own image.


blairdrummond:
If the R geospatial packages don't compile, then we can skip them and just use the previous subset

sylus:
This has been applied separately please take a look :D

PR #6, opened 2020-04-07
Split minimal image into base+minimal

blairdrummond:
`base` image is where the CPU v.s. GPU configuration happens

The minimal image is where the software stack gets added. Past this point, no one should really touch the base image, but minimal offers common software to all future packages.

There are a lot of commits because I had to do a rebase, but the diffs are pretty straight-forward. Some are large only because large blocks got copy-pasted between files.


sylus:
This is done now and hopefully will be the last re-arch, I pushed it off but had to make some additional changes.

Issue #7, opened 2020-04-07
Additional software


blairdrummond:
I don't want to mess up the pending pull request, so just putting the notes here. This cleans up most of the open issues here https://github.com/StatCan/terraform-kubernetes-aks-platform-daaas/issues

```
# Minimal or Base

## Issue #7
ENV PATH="/home/jovyan/.local/bin:${PATH}"

## Issue #9
pip install gpustat


## Look into this. Issue #5
conda install -c conda-forge jupyter_nbextensions_configurator

# Machine Learning Image

## Issue #10
conda install ... 
   'torchvision' \
   'torchtext' \

## Bayesian stuff; request from Ken.
pip install PyMC3

## XGBoost is missing
conda install -c conda-forge xgboost

```

sylus:
This is done now :D

PR #8, opened 2020-04-07
Issue #7 All tidied up. XGBoost and more

blairdrummond:
XGBoost, PyMC3 is a request I got from Ken, and then editing the PATH.


sylus:
The last build passed so merging this now with some slight tweaks :D

PR #21, opened 2020-04-27
jupyterlab update & golang & pachctl

### LINKED ISSUE #14: pyarrow
Might be something worth looking into? I will look into it. (Lower priority, obviously)

blairdrummond:
Added a couple R packages, pyarrow (closes #14), pachctl, and also installed golang.

@zachomedia  if you could take a look at the golang configuration and make suggestions that would be good. I'm a golang noob; hope that $GOPATH and $GOROOT are configured sensibly here.

### DIFF
--- a/minimal-notebook/cpu/Dockerfile
+++ b/minimal-notebook/cpu/Dockerfile
@@ -15,7 +15,7 @@ RUN apt-get update && \
     && \
     rm -rf /var/lib/apt/lists/*
 
-# kubectl, mc, az
+# kubectl, mc, az, pachctl
 RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt`/bin/linux/amd64/kubectl \
     && chmod +x ./kubectl \
     && sudo mv ./kubectl /usr/local/bin/kubectl \
@@ -77,46 +77,27 @@ RUN pip install --quiet \
     && \
     jupyter lab build && \
     jupyter lab clean && \
-    npm cache clean --force && \
-    rm -rf /home/$NB_USER/.cache/yarn && \
-    rm -rf /home/$NB_USER/.node-gyp && \
-    fix-permissions $CONDA_DIR && \
-    fix-permissions /home/$NB_USER
+  npm cache clean --force && \
+  rm -rf /home/$NB_USER/.cache/yarn && \
+  rm -rf /home/$NB_USER/.node-gyp && \
+  fix-permissions $CONDA_DIR && \
+  fix-permissions /home/$NB_USER
 
-# Doesn't work well yet, but it's in active development.
-# Retest in a few weeks.
-#RUN jupyter labextension install --no-build jupyterlab-s3-browser && \
-#    pip install jupyterlab-s3-browser && \
-#    jupyter serverextension enable --py jupyterlab_s3_browser && \
-#    jupyter lab build
-
-# Incompatible with current jupyterlab version.
-# Try these again later.
-#RUN conda install jupyter-archive flake8 autopep8
-#RUN conda install -c conda-forge python-language-server r-languageserver
-#RUN jupyter lab build
+# Solarized
+RUN mkdir -p /home/$NB_USER/.jupyter/lab/user-settings/@jupyterlab/apputils-extension && \
+  printf '{ "theme": "JupyterLab Solarized Dark" }\n' \
+  > /home/$NB_USER/.jupyter/lab/user-settings/@jupyterlab/apputils-extension/themes.jupyterlab-settings
 
 ENV GOROOT=/usr/local/go
-ENV GOPATH=/opt/go
 ENV PATH=$PATH:$GOROOT/bin
-
+ENV GOPATH=/home/$NB_USER/go/
 ARG LATEST_GO_VERSION=1.14.2
 RUN wget https://dl.google.com/go/go${LATEST_GO_VERSION}.linux-amd64.tar.gz -O /tmp/go.tar.gz \
-    && tar -xvf /tmp/go.tar.gz -C /usr/local \
-  && \
-    printf "\nexport GOROOT=$GOROOT" | tee -a /etc/profile && \
-    printf "\nexport GOPATH=$GOPATH" | tee -a /etc/profile && \
-    printf "\nexport PATH=\$GOPATH/bin:$GOROOT/bin:\$PATH\n" | tee -a /etc/profile \
- && \
-    printf "\nexport GOPATH=/home/$NB_USER/.go" | tee -a /home/$NB_USER/.bashrc && \
-    printf "\nexport PATH=\$GOPATH/bin:$GOPATH/bin:$GOROOT/bin:\$PATH\n" | tee -a /etc/profile /home/$NB_USER/.bashrc
-
-# Does not seem to work in non --privileged mode. Will revisit this option later...
-# RUN go get -v github.com/kahing/goofys && \
-#     go install -v github.com/kahing/goofys && \
-#     apt-get install -y fuse && \
-#     printf '\nuser_allow_other' >> /etc/fuse.conf && \
-#     rm -rf /var/lib/apt/lists/*
+  && tar -xvf /tmp/go.tar.gz -C /usr/local \
+  && printf "\nexport GOROOT=$GOROOT"             | tee -a /etc/profile \
+  && printf "\nexport PATH=\$GOROOT/bin:\$PATH\n" | tee -a /etc/profile \
+  && printf "\nexport GOPATH=$GOPATH\n"           | tee -a /etc/profile /home/$NB_USER/.bashrc \
+  && printf "\nexport PATH=$GOPATH/bin:\$PATH\n"  | tee -a /etc/profile /home/$NB_USER/.bashrc
 
 # Configure container startup
 EXPOSE 8888
--- a/minimal-notebook/cpu/Dockerfile
+++ b/minimal-notebook/cpu/Dockerfile
@@ -38,7 +38,7 @@ RUN pip install --quiet \
       'jupyter-lsp' \
     && \
     conda install --quiet --yes \
-    'jupyterlab=2.1.1' \
+    'jupyterlab>=2.1.1' \
     && \
     conda install --quiet --yes \
     -c conda-forge \
--- a/minimal-notebook/cpu/Dockerfile
+++ b/minimal-notebook/cpu/Dockerfile
@@ -37,12 +37,15 @@ RUN pip install --quiet \
       'jupyter-git' \
       'jupyter-lsp' \
     && \
+    conda update -n base conda && conda update --all && \
     conda install --quiet --yes \
     -c conda-forge \
       'ipywidgets' \
       'ipympl' \
       'jupyter_contrib_nbextensions' \
       'jupyterlab-dash' \
+      'jupyterlab-git' \
+      'xeus-python' \
       'nodejs' \
       'python-language-server' \
     && \
@@ -64,6 +67,12 @@ RUN pip install --quiet \
       '@lckr/jupyterlab_variableinspector' \
       'jupyter-matplotlib' \
       'jupyterlab-plotly' \
+      '@jupyterlab/debugger' \
+      '@jupyterlab/github' \
+      '@jupyterlab/git' \
+      'jupyterlab-theme-solarized-dark' \
+      'jupyterlab-execute-time' \
+      '@hadim/jupyter-archive' \
       'jupyterlab-spreadsheet' \
     && \
     jupyter lab build && \
--- a/minimal-notebook/cpu/Dockerfile
+++ b/minimal-notebook/cpu/Dockerfile
@@ -37,7 +37,9 @@ RUN pip install --quiet \
       'jupyter-git' \
       'jupyter-lsp' \
     && \
-    conda update -n base conda && conda update --all && \
+    conda install --quiet --yes \
+    'jupyterlab=2.1.1' \
+    && \
     conda install --quiet --yes \
     -c conda-forge \
       'ipywidgets' \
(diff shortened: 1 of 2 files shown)


minimal-notebook/cpu/Dockerfile
```diff
@@ -24,25 +24,37 @@ RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s
...
+    && sudo dpkg -i /tmp/pachctl.deb
+
 
 
 # Default environment
 RUN pip install --quiet \
       'jupyter-git' \
       'jupyter-lsp' \
     && \
+    conda update -n base conda && conda update --all && \
```
zachomedia, line 40: Please remove this, we don't want to be doing this in the docker image - it starts from fresh anyways.

minimal-notebook/cpu/Dockerfile
```diff
@@ -64,6 +83,41 @@ RUN pip install --quiet \
...
+#RUN jupyter labextension install --no-build jupyterlab-s3-browser && \
+#    pip install jupyterlab-s3-browser && \
+#    jupyter serverextension enable --py jupyterlab_s3_browser && \
+#    jupyter lab build
+
+# Incompatible with current jupyterlab version.
+# Try these again later.
+#RUN conda install jupyter-archive flake8 autopep8
+#RUN conda install -c conda-forge python-language-server r-languageserver
+#RUN jupyter lab build
```
zachomedia, line 97: Please remove all commented code from this PR, we would like to keep the Dockerfile clean. If you are actively working on it, it may be more appropriate in a separate branch.

minimal-notebook/cpu/Dockerfile
```diff
@@ -64,6 +83,41 @@ RUN pip install --quiet \
...
+
+ENV GOROOT=/usr/local/go
+ENV GOPATH=/opt/go
```
zachomedia, line 100: You shouldn't have to set `GOPATH`, it defaults to a directory in the user dir which is better than `/opt/go`

minimal-notebook/cpu/Dockerfile
```diff
@@ -64,6 +83,41 @@ RUN pip install --quiet \
...
+ && \
+    printf "\nexport GOPATH=/home/$NB_USER/.go" | tee -a /home/$NB_USER/.bashrc && \
+    printf "\nexport PATH=\$GOPATH/bin:$GOPATH/bin:$GOROOT/bin:\$PATH\n" | tee -a /etc/profile /home/$NB_USER/.bashrc
+
+# Does not seem to work in non --privileged mode. Will revisit this option later...
+# RUN go get -v github.com/kahing/goofys && \
+#     go install -v github.com/kahing/goofys && \
+#     apt-get install -y fuse && \
+#     printf '\nuser_allow_other' >> /etc/fuse.conf && \
+#     rm -rf /var/lib/apt/lists/*
```
zachomedia, line 119: Please remove commented code. If it's needed in the future, please place in its own branch.

minimal-notebook/cpu/Dockerfile
```diff
@@ -64,6 +83,41 @@ RUN pip install --quiet \
...
+ENV GOROOT=/usr/local/go
+ENV GOPATH=/opt/go
+ENV PATH=$PATH:$GOROOT/bin
+
+ARG LATEST_GO_VERSION=1.14.2
+RUN wget https://dl.google.com/go/go${LATEST_GO_VERSION}.linux-amd64.tar.gz -O /tmp/go.tar.gz \
+    && tar -xvf /tmp/go.tar.gz -C /usr/local \
+  && \
+    printf "\nexport GOROOT=$GOROOT" | tee -a /etc/profile && \
+    printf "\nexport GOPATH=$GOPATH" | tee -a /etc/profile && \
```
zachomedia, line 108: Same `GOPATH` comment

minimal-notebook/gpu/Dockerfile
```diff
@@ -33,17 +38,24 @@ RUN pip install --quiet \
       'jupyter-git' \
       'jupyter-lsp' \
     && \
+    conda update -n base conda && conda update --all && \
```
zachomedia, line 41: Same comment about remove `conda update`

minimal-notebook/gpu/Dockerfile
```diff
@@ -65,6 +84,41 @@ RUN pip install --quiet \
...
+#RUN jupyter labextension install --no-build jupyterlab-s3-browser && \
+#    pip install jupyterlab-s3-browser && \
+#    jupyter serverextension enable --py jupyterlab_s3_browser && \
+#    jupyter lab build
+
+# Incompatible with current jupyterlab version.
+# Try these again later.
+#RUN conda install jupyter-archive flake8 autopep8
+#RUN conda install -c conda-forge python-language-server r-languageserver
+#RUN jupyter lab build
```
zachomedia, line 98: Same comments comment.

minimal-notebook/gpu/Dockerfile
```diff
@@ -65,6 +84,41 @@ RUN pip install --quiet \
...
+    printf "\nexport GOPATH=/home/$NB_USER/.go" | tee -a /home/$NB_USER/.bashrc && \
+    printf "\nexport PATH=\$GOPATH/bin:$GOPATH/bin:$GOROOT/bin:\$PATH\n" | tee -a /etc/profile /home/$NB_USER/.bashrc
+
+# Does not seem to work in non --privileged mode. Will revisit this option later...
+# RUN go get -v github.com/kahing/goofys && \
+#     go install -v github.com/kahing/goofys && \
+#     apt-get install -y fuse && \
+#     printf '\nuser_allow_other' >> /etc/fuse.conf && \
+#     rm -rf /var/lib/apt/lists/*
+
```
zachomedia, line 121: Same comments comment

minimal-notebook/gpu/Dockerfile
```diff
@@ -65,6 +84,41 @@ RUN pip install --quiet \
...
+ENV GOROOT=/usr/local/go
+ENV GOPATH=/opt/go
+ENV PATH=$PATH:$GOROOT/bin
+
+ARG LATEST_GO_VERSION=1.14.2
+RUN wget https://dl.google.com/go/go${LATEST_GO_VERSION}.linux-amd64.tar.gz -O /tmp/go.tar.gz \
+    && tar -xvf /tmp/go.tar.gz -C /usr/local \
+  && \
+    printf "\nexport GOROOT=$GOROOT" | tee -a /etc/profile && \
+    printf "\nexport GOPATH=$GOPATH" | tee -a /etc/profile && \
```
zachomedia, line 109: So `GOPATH` comment

minimal-notebook/gpu/Dockerfile
```diff
@@ -65,6 +84,41 @@ RUN pip install --quiet \
...
+#    jupyter lab build
+
+# Incompatible with current jupyterlab version.
+# Try these again later.
+#RUN conda install jupyter-archive flake8 autopep8
+#RUN conda install -c conda-forge python-language-server r-languageserver
+#RUN jupyter lab build
+
+ENV GOROOT=/usr/local/go
+ENV GOPATH=/opt/go
```
zachomedia, line 101: So `GOPATH` comment

minimal-notebook/cpu/Dockerfile
```diff
@@ -64,6 +83,41 @@ RUN pip install --quiet \
...
+#    jupyter serverextension enable --py jupyterlab_s3_browser && \
+#    jupyter lab build
+
+# Incompatible with current jupyterlab version.
+# Try these again later.
+#RUN conda install jupyter-archive flake8 autopep8
+#RUN conda install -c conda-forge python-language-server r-languageserver
+#RUN jupyter lab build
+
+ENV GOROOT=/usr/local/go
```
zachomedia, line 99: I don't think you have to set `GOROOT`, it usually defaults to the right value.

minimal-notebook/gpu/Dockerfile
```diff
@@ -65,6 +84,41 @@ RUN pip install --quiet \
...
+#    jupyter serverextension enable --py jupyterlab_s3_browser && \
+#    jupyter lab build
+
+# Incompatible with current jupyterlab version.
+# Try these again later.
+#RUN conda install jupyter-archive flake8 autopep8
+#RUN conda install -c conda-forge python-language-server r-languageserver
+#RUN jupyter lab build
+
+ENV GOROOT=/usr/local/go
```
zachomedia, line 100: I don't think you have to set `GOROOT`, it usually defaults to the right value.

minimal-notebook/cpu/Dockerfile
blairdrummond: Unfortunately it's necessary. The latest version of conda is 4.8.3, even though they only have installers for 4.8.2 (latest = 4.8.2 by the checksum)

https://repo.anaconda.com/miniconda/

minimal-notebook/cpu/Dockerfile
blairdrummond: Ok sure thing

minimal-notebook/cpu/Dockerfile
blairdrummond: Ok

minimal-notebook/cpu/Dockerfile
blairdrummond: To clarify, I mean that the images that we pull are `4.8.2`, and many extensions don't work as a result. Going up to 4.8.3 was the only way to get git and a couple other extensions working

minimal-notebook/cpu/Dockerfile
blairdrummond: I don't think that's true in this installation; I just unpacked the tarball to `/usr/local/go`, so I don't see how it could know

minimal-notebook/cpu/Dockerfile
```diff
@@ -55,14 +67,37 @@ RUN pip install --quiet \
...
+  fix-permissions /home/$NB_USER
+
+# Solarized
+RUN mkdir -p /home/$NB_USER/.jupyter/lab/user-settings/@jupyterlab/apputils-extension && \
+  printf '{ "theme": "JupyterLab Solarized Dark" }\n' \
+  > /home/$NB_USER/.jupyter/lab/user-settings/@jupyterlab/apputils-extension/themes.jupyterlab-settings
+
+ENV GOROOT=/usr/local/go
+ENV PATH=$PATH:$GOROOT/bin
+ENV GOPATH=/home/$NB_USER/go/
```
blairdrummond, line 93: I found that `$GOPATH` was not set if I didn't include this. While it would default correctly without it being set, the path to any go installed programs would be broken.

minimal-notebook/cpu/Dockerfile
zachomedia: I'm not sure this makes sense to me. Especially since 4.8.3 is just a minor update, shouldn't have any impactful changes like that? Can we get the base updated to 4.8.3 rather than updating it mid-build

blairdrummond: Unfortunately it does. The extensions like `git` and debugging *simply do not work* on 4.8.2. I looked at updating the other images but it's complicated because of the way that the GPU image does it (and the fact that the Jupyterlab datascience-notebook seems to ship with 4.8.2).

The issue is that there is no way to make this work with 4.8.3

https://github.com/StatCan/kubeflow-containers/blob/7cd5b08f171f156da4da2a9f57df6587a3bc8512/base-notebook/gpu/Dockerfile#L83-L89

They don't have an analogous installer :-/

blairdrummond: So we could move the changes over to base, but it will just be running the same conda update then cleaning the conda cache. I don't think we can install 4.8.3 from the beginning. I think we might have to wait until a new copy of the installer comes out.

blairdrummond: Actually, it might be that the issue isn't conda, but just that doing the update --all upticked the Jupyterlab from 2.0.1 to 2.1.1 and that's what's fixing things. That makes more sense.

blairdrummond: @zachomedia So how about I see if I can remove the conda update and just update jupyterlab earlier? I can update jupyterlab at the source in the GPU image, but I have to add a line to manually update it in the cpu base image.

blairdrummond: Ok so it can be done, but it's tricky and requires adding an additional `jupyterlab build` step. This would be the result `base-cpu` image

```
FROM jupyter/datascience-notebook
USER root
ENV PATH="/home/jovyan/.local/bin:${PATH}"

# this block is new.
RUN conda install --quiet --yes \
    'notebook=6.0.3' \
    'jupyterhub=1.1.0' \
    'jupyterlab=2.1.1' && \
    jupyter lab build && \
    jupyter lab clean && \
    conda clean --all -f -y && \
    npm cache clean --force && \
    rm -rf $CONDA_DIR/share/jupyter/lab/staging && \
    rm -rf /home/$NB_USER/.cache/yarn && \
    fix-permissions $CONDA_DIR && \
    fix-permissions /home/$NB_USER

... # more stuff

# Configure container startup
EXPOSE 8888
USER jovyan
COPY start-custom.sh /usr/local/bin/
ENTRYPOINT ["tini", "--"]
CMD ["start-custom.sh"]
```

The GPU image would be easier to edit because it already has a block like the one added here.

Would this route be preferable? It means an extra `jupyter build` step over what we have now

blairdrummond: I replaced the `conda update --all` with just updating jupyterlab to the needed version. It could also be done in the base image, but it creates an extra `jupyter lab build` step which we don't need if we bundle it with the rest of the jupyterlab configuration.

zachomedia:
This seems reasonable to me now, @blairdrummond is this ready to be merged?

blairdrummond:
Yep, it's good to merge. Thanks!

PR #24, opened 2020-04-30
feat(notebook): Non-root CPU ml-workspace with RStudio

### LINKED ISSUE #18: request for RStudio
Hello,

With ml-workspace being enabled it would be great have RStudio as a default installation on the images.

Thank you.

### LINKED ISSUE #28: ml-workspace: non-root user
By default, ml-workspace images have everything hardcoded to /root and running as root. This issue tracks changing that, starting with refactoring to a non-root single user ('joyvan').

[Upstream issue](https://github.com/ml-tooling/ml-workspace/issues/11)

PR for initial attempt (cpu-only due to image size issues discussed in #26):  #24

frazs:
Working but huge (16.7GB), I will need to push this manually to k8scc01covidacr.azurecr.io/ml-workspace-cpu once approved (with the commit SHA as the tag as Brendan requested unless there is a better way to go). 

The tools scripts have been changed. This is because they are installed with sudo, but in their original form they could also run the program they install, which would be as root and thus not great for security nor user settings. The code to run the tools has been removed.

I am not yet fully confident that this is the minimal set of elevated permissions / changed ownership for the image to run, but I can experiment further, and it is at the very least an improvement on all-root everything. Please let me know if anything stands out as a particular security concern.

As nearly all of the image size bloat is due to docker layer file duplication on chown and chmod, particually for /opt/conda the less that can be modified the better. But there may be other avenues (e.g. chown/chmod on image initialization instead of built-in, perhaps through an exec script or another method, which could also potentially support dynamic/multiple users if ever needed as opposed to everyone as joyvan).

Resolves #18 
Resolves #28


blairdrummond:
@frazs Should we implement the size reduction before the push? What do you think?

frazs:
I think it's alright to push without the size reduction for now. The ~6gb size difference on the CPU image wouldn't make a major impact at a small scale, and the reduction will most likely be ready before there are a lot more users.

brendangadd:
@frazs Merging this should also close #28, right? (If so, please add to linked issues.)

frazs:
I cannot merge this in. Requesting review from anyone who can. Work is continuing on a separate image based on the raw Dockerfile, but any reported issues will be handled there and may be possible to patch in here.

brendangadd:
@frazs: I changed the base (destination) branch this will be merged to. We'll protect that branch so that this work isn't lost. Your future work to produced a layered non-root build with a base image is what will end up in `master`.

PR #46, opened 2020-05-23
feat: add R-Studio image

blairdrummond:
Image is built off of jupyter/r-notebook. It supports `/tree` view,
but by default launches an rstudio interface.

The image comes with R libraries (like dplyr), arrow, odbc+sqlite.


r-studio/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+  && \
+    wget https://dl.min.io/client/mc/release/linux-amd64/mc \
+    && chmod +x mc \
+    && mv mc /usr/local/bin/mc \
+  && \
+    sh -c 'curl -sL https://aka.ms/InstallAzureCLIDeb | bash' \
+  && \
+    curl -o /tmp/pachctl.deb -L \
+      https://github.com/pachyderm/pachyderm/releases/download/v1.10.0/pachctl_1.10.0_amd64.deb \
+    && sudo dpkg -i /tmp/pachctl.deb
```
brendangadd, line 79: I know this is coming from straight from base, but let's checksum. Can do separate issue for base image.

sylus:
Can we keep this cleaner it looks like there is notebooks attached? This repo should only be for Dockerfile and related scripts.

sylus:
Also I think we should be inherited from base like all of the others if need something from r-notebook it should only be minimial lines. I can help with this.

Sorry not trying to be a pain just want to minimize any future operational concerns.

blairdrummond:
@sylus  No problem, I can pull the notebooks from a repo. What about

https://github.com/StatCan/daaas/issues/82

You can't (easily) read ipynb files from RStudio, so I think a different repo is warranted for R-Studio

As for using the common base image, I can do that but it will increase the image size (JupyterLab is big). Current image is 5.5gb, using a different base (like minimal) will probably crank that up to 8gb or so. I tested that out here

https://github.com/StatCan/kubeflow-containers/issues/32

I'm ok with either, just listing the pro/cons

brendangadd:
> Also I think we should be inherited from base like all of the others if need something from r-notebook it should only be minimial lines. I can help with this.

I can see why @blairdrummond didn't inherit from our base image; it inherits from deep in the core stacks tree and comes with a bunch of software that's extraneous in this context. By inheriting from r-notebook, this image backs out of two layers worth of Python analysis tools.

r-studio/Dockerfile
blairdrummond: Registered as issue #47

brendangadd:
* StatCan/daaas#82 will move examples into a separate repo and update this image to clone it
* #47 will checksum software downloaded from external sources (plus the base image)
* #48 will take a look at the broader architectural concerns of two sources of inheritance for layering our images

This image is a great first step. With outstanding concerns tracked as per above, LGTM.

PR #49, opened 2020-05-27
fix: remove --allow-root option

### LINKED ISSUE #31: Remove `--allow-root` ?
Um. I think we might not even need --allow-root in the kubeflow images. I removed it and launched the notebook just fine.

try out this image
<the registry>/kf-notebook-no-root

Or build it yourself

no-root-test.tar.gz

@brendangadd @sylus @ca-scribner

Any thoughts on what we'd need to test before removing things? Based on the notes in

kubeflow/kubeflow#300

it seems like they left --allow-root in so that uses could apt-get?

blairdrummond:
Closes #31 . This --allow-root option seems unnecessary, as we don't
run as root, and the option only seems to be suggested for the that
case.

See https://github.com/kubeflow/kubeflow/issues/300 too.

### DIFF
--- a/base-notebook/cpu/start-custom.sh
+++ b/base-notebook/cpu/start-custom.sh
@@ -9,7 +9,6 @@ git clone https://github.com/statcan/jupyter-notebooks
 jupyter notebook --notebook-dir=/home/${NB_USER} \
                  --ip=0.0.0.0 \
                  --no-browser \
-                 --allow-root \
                  --port=8888 \
                  --NotebookApp.token='' \
                  --NotebookApp.password='' \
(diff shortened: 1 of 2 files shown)


sylus:
Are we sure this is ok? It is in both the kubeflow documentation and in all of their notebooks:

https://github.com/kubeflow/kubeflow/blob/06bf3a40679a64cecb463c7a6363c678e5ba42a9/components/tensorflow-notebook-image/Dockerfile#L126

blairdrummond:
I've run the images without it no problem. In fact the current R-Studio image does not have `--allow-root`. I haven't encountered any issues.

I think that the option is just an artifact from usage by root.

I think there's no harm in leaving it in though, either. The containers themselves aren't running as root anyway, so I suppose there's no harm?

sylus:
@zachomedia can u look at this will this cause regressions?

blairdrummond:
Found my comment where I looked at his previously. In the comment is a notebook server image which doesn't use --allow-root

https://github.com/StatCan/daaas/issues/47#issuecomment-627546201

brendangadd:
Checked in with @zachomedia, who does not foresee any issues with this.

PR #50, opened 2020-05-27
refactor: remove ml-workspace image

blairdrummond:
ML-Workspace is now in its own repo, so it's not needed here.


sylus:
Did this manually as did a few changes earlier ^_^

PR #64, opened 2020-06-20
WIP: feat(checksums): add a checksum gathering tool

blairdrummond:
Add a checksum gathering tool, for keeping versions and sha256sums up-to-date


blairdrummond:
Would love a quick review of the methodology! With an OK here, I'll merge in the checksums. (Ignoring `base/gpu` since those versions are pinned but come from a fork)

justbert:
And this tool will be used to gather checksums for new versions?

blairdrummond:
Yep exactly. I figured we'd want some more automated way to detect when new releases of pinned programs come in, to make it easier to stay up-to-date (without manually reading through the dockerfiles and checking every pinned package to see if it's current). 

It generates a diff against the CHECKSUMS file, so that it should be able to report when something has been updated. Could probably be incorporated into a CI process if we want?

justbert:
Technically, the kubectl version should be +-1 version away from the cluster you want to manage.

blairdrummond:
Good point! I did not think about that...

justbert:
I guess I'm still confused as to how this will be implemented in the Dockerfile.

blairdrummond:
Yeah I was thinking just information gathering. Want to make maintenance easier, and provide automated alerts when maintenance might be required, but I think better not to take on automating the Docker images.

justbert:
As long people don't become overzealous and update everything causing breaks in the images! :)

blairdrummond:
```
# Dockerfile
FROM archlinux
...
```

blairdrummond:
I am hoping to avoid going through the process of building this myself (not sure if my machine can take it) by waiting for #65  first. If that goes through, then should be able to easily test out the build inside of this PR, and then it should be good-to-go (assuming there are no other comments).

blairdrummond:
Closing to move to a feature-branch

PR #65, opened 2020-06-21
feat(ci/cd): add a build (no push) step to PRs

blairdrummond:
Would be great to be able to test builds inside of Github. (Some of these images are so large they occasionally crash my personal computer).

This PR moves the old `build.yml` to `publish.yml` (since it builds and pushes) and created a new `build.yml` which *just* builds. (This is consistent with [daaas-containers](https://github.com/StatCan/daaas-containers) )

**Note: The images built in the PR v.s. on master will not be identical because on master, parent images will be mutated by the build process, and in PRs they will not**.

However, I think there are still merits to having *some* build process CI, even if there is a limitation here in terms of not capturing inter-image software changes.


sylus:
I think this makes sense.

@zachomedia will need you to +1 this

PR #67, opened 2020-06-22
fix(build.yml): move the docker prune till after scanning

blairdrummond:
I think this PR (#64 ) Is failing because the image gets pruned before it can be scanned. This isn't a problem in the publish image because it can grab the image from the registry, but it's an issue for `build.yml`

@sylus fixed this in daaas-containers a few days ago, and basically just copying the fix from there

https://github.com/StatCan/daaas-containers/blob/master/.github/workflows/build.yml

Just needed to switch the prune command around.


blairdrummond:
Thanks @sylus !

PR #71, opened 2020-06-26
feat(rstudio): update r-catools and r-bitops

frazs:
Resolves https://github.com/StatCan/kubeflow-containers-desktop/issues/16

Note: I also added `RUN chmod +x /usr/local/bin/start-custom.sh` as otherwise, both locally and on Kubeflow, the container would not start: `[FATAL tini (6)] exec start-custom.sh failed: Permission denied`. Not sure what's up with that if it was working before.

### DIFF
--- a/r-studio/Dockerfile
+++ b/r-studio/Dockerfile
@@ -20,7 +20,6 @@ RUN python3 -m pip install \
     conda install --quiet --yes \
       'r-arrow' \
       'r-aws.s3' \
-      'r-catools' \
       'r-devtools' \
       'r-dplyr' \
       'r-ggplot2' \
@@ -33,6 +32,11 @@ RUN python3 -m pip install \
       'r-shiny' \
       'r-sparklyr' \
       'r-tidyr' \
+    && \
+    # Get the latest r-catools and r-bitops from conda-forge
+    conda install -c conda-forge --quiet --yes \
+     'r-catools' \
+     'r-bitops' \
     && \
     conda clean --all -f -y && \
     fix-permissions $CONDA_DIR && \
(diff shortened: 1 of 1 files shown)


r-studio/Dockerfile
```diff
@@ -34,6 +33,11 @@ RUN python3 -m pip install \
       'r-sparklyr' \
       'r-tidyr' \
     && \
```
blairdrummond, line 35: Could we add this within the previous conda install block? My bet is that it's the same. (Not that it really changes anything)

r-studio/Dockerfile
```diff
@@ -82,6 +86,7 @@ RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s
 
 
 COPY start-custom.sh /usr/local/bin
+RUN chmod +x /usr/local/bin/start-custom.sh
```
blairdrummond, line 89: Doesn't hurt! I think the permissions *should* match what's in your directory. Were you building on a *Nix system?

https://stackoverflow.com/questions/34445604/docker-copy-command-give-777-access-to-the-copied-file

sylus: Do we really need this? I really don't like that rstudio has a different base then all the others it makes maintainin more difficult

blairdrummond: I am ok with rebasing it onto either minimal or base; the image will be relatively large tho

frazs:
Whoops, actually none of this is necessary. I had made an incorrect assumption that this Dockerfile reflected what is up on Kubeflow, so I thought that `r-catools` must have been pulled from a less up-to-date channel than conda-forge, and I should have first verified that by going into the image and doing a `conda list`.

In fact, what is up on Kubeflow is the [version from May 26](https://github.com/StatCan/kubeflow-containers/blob/937eb8bf35fae5cd303b8d02c5bd7fd71b270fd7/r-studio/Dockerfile) that *doesn't install `r-catools` at all*.

As for `r-bitops`, installing it is not necessary when `r-catools` has been installed, as the relevant packages are already pulled in by `r-catools`. I do not see any errors on creating an R notebook.

We can resolve the issue by updating the dropdown to match the latest version of this Dockerfile, and make a separate issue for rebasing it.

PR #72, opened 2020-06-30
feat(elyra): Back to stable

sylus:
Just switching this back to stable since Elyra now supports the newer jupyterlab and doesn't need my fork anymore.

Also enabled the the extension the code snippets package.


justbert:
LGTM! Though I'm not super familiar with these components!

sylus:
Thank you :)

PR #82, opened 2020-07-28
refactor(r-studio): rebase r-studio to minimal-notebook-cpu

saffaalvi:
Resolves https://github.com/StatCan/kubeflow-containers/issues/75

Rebased `r-studio` to extend `minimal-notebook-cpu` and installed missing packages (libssl1.0.0.deb and libssl1.0.2.deb) into `r-studio`.


r-studio/Dockerfile
```diff
@@ -1,10 +1,14 @@
...
 USER root
 
 # install rstudio-server
 ARG RSTUDIO_VERSION=1.1.463
 ARG SHA256=62aafd46f79705ca5db9c629ce3b60bf708d81c06a6b86cc4b417fbaf30691c1
 RUN apt-get update && \
+    wget http://archive.ubuntu.com/ubuntu/pool/main/o/openssl1.0/libssl1.0.0_1.0.2n-1ubuntu5.3_amd64.deb -O libssl1.0.0.deb && \
+    dpkg -i libssl1.0.0.deb && \
+    wget http://ftp.br.debian.org/debian/pool/main/o/openssl1.0/libssl1.0.2_1.0.2u-1~deb9u1_amd64.deb -O libssl1.0.2.deb && \
+    dpkg -i libssl1.0.2.deb && \
```
frazs, line 11: Are both of these necessary or would one of them work? I'm wondering if the libssl1.0.2 ends up overwriting the libssl1.0.0

minimal-notebook/cpu/Dockerfile
```diff
@@ -103,7 +103,7 @@ RUN pip install --quiet \
       'jupyterlab-spreadsheet' \
       'nbdime-jupyterlab' \
     && \
-    jupyter lab build && \
+    jupyter lab build --dev-build=False --minimize=False && \
```
frazs, line 106: Why was this changed?

frazs:
`base-notebook/cpu/Dockerfile` and `minimal-notebook/cpu/Dockerfile` are reporting `No newline at end of file`. @brendangadd you've fixed this in the past (e.g. https://github.com/StatCan/kubeflow-containers-desktop/commit/b81ffd7b36e9884871c666cf3ffa26b3c19b2c43), how do you do that?

minimal-notebook/cpu/Dockerfile
saffaalvi: When trying to build the `minimal-notebook-cpu` image locally, for some reason, I kept getting an error and this fix was recommended on the terminal. Once I changed this line, the error was resolved.

r-studio/Dockerfile
saffaalvi: When libssl1.0.0 is removed, the image will build successfully but when the container is ran the following error comes up in the terminal: `rserver: error while loading shared libraries: libssl.so.1.0.0: cannot open shared object file: No such file or directory` and then when the container is opened in the browser, this comes up:

error

I just tried removing libssl1.0.2 and no errors come up in the image or container, so we can actually remove libssl1.0.2 and keep libssl1.0.0.

minimal-notebook/cpu/Dockerfile
frazs: I recommend leaving that as a local change and not committing it for now, because I am not sure how this would interact with the automated build process that takes place after merging. It is good to keep in mind though in case it occurs elsewhere.

sylus:
Any chance to get this in today? Would be awesome :D Think latest commit is missing what blair wanted added.

saffaalvi:
> Blair suggests adding these to r-studio/Dockerfile: https://github.com/jupyter/docker-stacks/blob/master/r-notebook/Dockerfile#L26-L52

Added the new packages to r-studio/Dockerfile just without specific versions as this was causing errors when building the image.

sylus:
Can fix the whitespace and new line issues?

PR #113, opened 2020-09-10
Remove prune from build CI

ca-scribner:
WIP testing CI.  Do not merge as is


sylus:
I think if you remove the prune you will hit CI limits.

PR #115, opened 2020-09-11
remove dremio and pachctl references

saffaalvi:
Resolves https://github.com/StatCan/kubeflow-containers/issues/105


base-notebook/cpu/Dockerfile
```diff
@@ -19,29 +19,16 @@ RUN pip --no-cache-dir install --quiet \
       'fire==0.3.1' \
       'git+https://github.com/kubeflow/pipelines@1d86111d8f152d3ed7506ea59cee1bfbc28abbf9#egg=kfp-azure-databricks&subdirectory=samples/contrib/azure-samples/kfp-azure-databricks'
 
-ARG DREMIO_VERSION=1.4.2.1003
-ARG DREMIO_FULL_VERSION=1.4.2.1003-1
-ARG DREMIO_RPM=https://download.dremio.com/odbc-driver/${DREMIO_VERSION}/dremio-odbc-${DREMIO_FULL_VERSION}.x86_64.rpm
-ARG DREMIO_SHA=d2f5ff5d498d97f29a0eedc485dbcc85faf0e82eed2c1f7eb0ef3fee393ad450
-
-# Added ODBC Support for Dremio
 RUN conda install --quiet --yes pyodbc && \
```
ca-scribner, line 22: do we need the rest of this ODBC stuff or was it only here for Dremio?

base-notebook/gpu/Dockerfile
```diff
@@ -380,29 +380,16 @@ RUN pip --no-cache-dir install --quiet \
       'fire==0.3.1' \
       'git+https://github.com/kubeflow/pipelines@1d86111d8f152d3ed7506ea59cee1bfbc28abbf9#egg=kfp-azure-databricks&subdirectory=samples/contrib/azure-samples/kfp-azure-databricks'
 
-ARG DREMIO_VERSION=1.4.2.1003
-ARG DREMIO_FULL_VERSION=1.4.2.1003-1
-ARG DREMIO_RPM=https://download.dremio.com/odbc-driver/${DREMIO_VERSION}/dremio-odbc-${DREMIO_FULL_VERSION}.x86_64.rpm
-ARG DREMIO_SHA=d2f5ff5d498d97f29a0eedc485dbcc85faf0e82eed2c1f7eb0ef3fee393ad450
-
-# Added ODBC Support for Dremio
 RUN conda install --quiet --yes pyodbc && \
```
ca-scribner, line 383: Same comment as above

ca-scribner (changes requested):
might be able to remove a few more things.  Not sure if they're used elsewhere though

base-notebook/cpu/Dockerfile
blairdrummond: I think ODBC is a good idea in case someone wants to connect to an external database. Was needed for Dremio, but I think it's a good idea to have in general.

ca-scribner: I think the ODBC stuff is ~1/2 the total build time for this (just tested locally and assuming it'll be the same online) - adds ~8 min per build.  

To connect to other db's I think users only need to `pip install pyodbc` right?  If so, I vote we remove to keep the rebuilds shorter and let users install when they need it.  @blairdrummond?

ca-scribner: hmm users can't `apt-get install unixodbc-dev` for themselves though and that's a requirement on linux.  

I think most of the slowness here comes from conda - we could `apt-get install unixodbc-dev` then `pip install pyodbc` instead as a compromise

ca-scribner (commented):
Looks good to me

ca-scribner (approved):

PR #124, opened 2020-09-18
fix: update ubuntu version

saffaalvi:
Resolves https://github.com/StatCan/kubeflow-containers/issues/123


blairdrummond:
Looks good to me! I'm sure there is another long-term fix but seems like a good idea to fix the build right away

blairdrummond (approved):

Issue #48, opened 2020-05-26
Improve image inheritance strategy


brendangadd:
We want to benefit as much as possible from upstream fixes in Jupyter's core docker stacks, but also be able to inherit from our own base image so we can fix our own layers in a minimal number of locations. Right now, our base image inherits from deep in the core docker stack tree.

I'm wondering if we should use a mixin approach so we can benefit from upstream fixes to the whole tree without pushing our base image to a leaf node. This will allow us to extend from multiple locations in the core tree while preserving common config of our own.

sylus:
I think this issue was closed with the recent merge?

Feel free to re-open but this looks solved to me.

PR #129, opened 2020-09-23
feat: Merge the CPU and GPU Dockerfiles into shared definitions

### LINKED ISSUE #107: Refactor notebook images to reduce duplication
Most of our CPU and GPU notebook stack is identical, with the difference being the `FROM` line at the top of the file (and some inconsistencies that can be ironed out).  We also have some duplication between our stack and the upstream images (we reinstall some things that they also installed.  Not sure if this was intentional to update versions or accidental).  The CPU/GPU notebooks are also a little out of sync (at least they appear to be), with a few things being installed only in one.

Refactor our notebook Dockerfiles to reduce duplication (between CPU/GPU and between ours and our upstream).  Also refactor/document to make it clear what we intentionally duplicate.  A possible solution would be to use an `ARG` for base image, then CI can pass the CPU or GPU base image at build time.  

Our CPU and GPU base images are also very different because the CPU inherits a lot from upstream whereas the GPU starts nearly from scratch and has to install the equivalent software to the CPU image.  Maybe we should break base-gpu into two images, one that recreates everything that the CPU upstream image gives us (so it's product is a GPU equivalent to the upstream CPU image), then we can combine what remains into a single cpu/gpu base image which takes upstream image as an `ARG`.

Adding a flow chart to the readme.md to show our inheritance would also be helpful for future maintainers.

### LINKED ISSUE #122: Tensorflow missing cuDNN drivers
Our GPU dockerfile stack is rooted on an NVIDIA image (`nvidia/cuda:10.1-base-ubuntu18.04`) which does not include the CUDA Deep Neural Network libraries which tensorflow requires.  This is discussed in iot-salzburg/gpu-jupyter#25.

iot-salzburg/gpu-jupyter#25 resolves the issue by basing off a different NVIDIA image - using the current release of [iot-salzburg/gpu-jupyter](https://github.com/iot-salzburg/gpu-jupyter) (or changing our NVIDIA base) would resolve this issue.  

Current work here on #114 has a solution mocked up for this issue

ca-scribner:
feat: Merge the CPU and GPU Dockerfiles into shared definitions

The CPU and GPU Dockerfile streams are virtually identical except for their upstream images.  This duplication increases maintenance efforts and the chance of misalignment between cpu and gpu machines (minor accidental differences in versions/tools exist in the released images).  This PR merges these two streams into a single Dockerfile to address these issues.  The requirement that CPU and GPU images base off different upstream images is addressed by accepting the `BASE_CONTAINER` as a `docker build` `--build-arg` (eg, `docker build --build-arg some-docker-stacks-image` for cpu, `docker build --build-arg some-gpu-jupyter-image` for gpu.

Resolves #107 
Resolves #122 

Todo before merge:
* [ ] remove the CI trigger for pull requests (set it back to just triggering on push to master)
* [ ] squash

Combine cpu/gpu images
-	Edit Dockerfiles to accept a build-arg BASE_CONTAINER for the upstream `FROM` image so that cpu and gpu builds can share a Dockerfile
-	Add `upstream-equivalent-notebook-gpu` to create the GPU version of the upstream image for `base-notebook`.  This pulls `gpu-jupyter` to create the GPU Dockerfile then builds that Dockerfile
-	Resolve any unintended differences between cpu/gpu images
-	Remove cpu/gpu subdirs

Update CI
-	Add version pinning during CI by passing `upstream_image:this_sha` rather than `upstream-image:latest` (ex: to build `minimal-notebook` we pass `base-notebook:this_sha` as the upstream image. This ensures that no cross-talk can occur between two concurrent builds from different tasks
-	Remove `docker system prune -f -a` from each CI image build step.  This caused each build step to re-pull the previous step's layers.  `build_push.sh` can optionally prune only recent layers if space saving is required
-	Add layer caching to the CI to speed up deployment/testing.  CI now does `docker pull this_image:master` prior to each `docker build` in order to pull the most recently built image in an attempt to get layers that `docker build` can cache from.  If a perfect cache hit, each image builds in <5 min instead of 15-35 min.  This also prevents rebuilding images unnecessarily (eg: if making a CPU-only change, the GPU images will just pull from cache rather than rebuild). In the worst case (editing `base-notebook`) this adds ~3-5min to build.
-	Encapsulated the typical build logic for each image (pull for cache, build image, tag latest/sha, push, optionally clean) into `scripts/build_push.sh`
-	Add `build_settings.env` to document upstream images for CI and local dev

Add documentation
-	expand top level readme.md
-	add flowcharts showing inheritence 
-	add readme.md in each image subdir

Add doc/convenience scripts for developers
-	`build*.sh` scripts added to root and image subdirs to help automate development

Misc fixes:
-	add `fix-permissions $CONDA_DIR && fix-permissions /home/$NB_USER` to a few `pip install` and `conda install`s that were missing them (resulting in image size).  Add a note to readme.md about this

Future work:
-	remove debug code from CI
-	pytorch install includes install for cuda drivers.  Does this break the cpu-only pytorch?  If yes, add build-arg for cpu/gpu toggle?


ca-scribner:
resolves #48

PR #132, opened 2020-09-29
feat: updated inherited image for cpu images

saffaalvi:
Resolves https://github.com/StatCan/kubeflow-containers/issues/117 and https://github.com/StatCan/kubeflow-containers/issues/130

* updated jupyter/datascience-notebook image version (python 3.8.5)
* updated package versions
* installed pytorch/torch using pip instead of conda
* removed double package installations

### DIFF
--- a/machine-learning-notebook/cpu/Dockerfile
+++ b/machine-learning-notebook/cpu/Dockerfile
@@ -4,8 +4,8 @@ USER root
 # Install Tensorflow
 RUN conda config --set channel_priority false && \
     conda install --quiet --yes \
-      'tensorflow=2.1*' \
-      'keras==2.3.1' \
+      'tensorflow==2.2.0' \
+      'keras==2.4.3' \
     && \
     conda clean --all -f -y && \
     fix-permissions $CONDA_DIR && \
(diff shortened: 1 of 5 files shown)


machine-learning-notebook/cpu/Dockerfile
```diff
@@ -4,8 +4,7 @@ USER root
 # Install Tensorflow
 RUN conda config --set channel_priority false && \
     conda install --quiet --yes \
-      'tensorflow=2.1*' \
-      'keras==2.3.1' \
+      'keras==2.4.3' \
```
ca-scribner, line 7: Since we're removing tensorflow here only because it is already installed as a byproduct of kubeflow-metadata (via ml-metadata), can we explicitly install tensorflow in base-notebook and put a note saying why (that it will be installed by kubeflow-metadata anyway until [link issue] is resolved)?  Then it is a bit clearer why its up front and could even leave it commented here or a note saying see base-notebook.

Just worried if we don't do something, we're going to suddenly lose tensorflow as soon as ml-metadata gets rid of their tf dependency and we won't notice why

ca-scribner (changes requested):
Please make sure it is explicit in the final commit message why tensorflow and kubernetes are removed (say they're installed earlier and cite by what command)

machine-learning-notebook/cpu/Dockerfile
saffaalvi: tensorflow is now pinned in minimal-notebook/cpu

ca-scribner (approved):

PR #145, opened 2020-11-03
Short hashes

Nancy-Badran:
Resolves StatCan/daaas#257


.github/workflows/build-cpu.yml
```diff
@@ -14,6 +14,7 @@ name: build cpu
 on:
   push:
     branches:
+      - Short-Haches
```
sylus, line 17: Can we remove this branch, it should only happen on master? Assume just for testing?

Also is spelled wrong, should be hashes

Nancy-Badran: Yeah, I was testing the short sha ,

brendangadd (changes requested):
Let's use eight characters.
* `git rev-parse --short` defaults to seven
* GitLab uses eight

sylus:
I think this needs to be rebased from latest?

Nancy-Badran:
Yeah @sylus , I will rebase it from latest, thanks

Nancy-Badran:
workflow changed, proposed changes in PR are out of date

PR #151, opened 2020-11-09
Combine base notebook cpu gpu images

JessicaBarh:
This Relates to PR [#129](https://github.com/StatCan/kubeflow-containers/pull/129) that is being split up into smaller PRs for better review. 

Resolves part of: https://github.com/StatCan/kubeflow-containers/issues/146 

Refactored base-notebook-gpu. upstream-equivalent-notebook-gpu will be built on the fly by this CI from a SHA-pinned pull of GPU-Jupyter to create the GPU equivalent to docker-stacks' datascience-notebook, which is our CPU upstream image. This lets both CPU and GPU versions of base-notebook be identical except for the upstream image.


sylus:
@zachomedia you are going to have to review this as can't quite make sense of it.

.github/workflows/build-cpu.yml
```diff
@@ -15,12 +15,17 @@ on:
   push:
     branches:
       - master
-
+  pull_request:
+    branches: 
+      - master  # DEBUG
```
ca-scribner, line 20: I like the idea of leaving this here.  It'll let people test in CI without actually merging into master.  The images are pushed with branch or commit ID, so it shouldn't disrupt anything.  @zachomedia that ok for you?

If we leave it, we can remove the `# DEBUG` comment

.github/workflows/build-cpu.yml
```diff
@@ -41,14 +46,28 @@ jobs:
         sudo apt clean
         docker rmi $(docker image ls -aq)
         df -h
-
+    
+    # tests if GITHUB_HEAD_REF is non-zero in length, then assigns the branch name to $BRANCH_NAME
+    - name: Get branch name for image tag
+      id: get_branch_name
+      run: |
+        if [ -z "$GITHUB_HEAD_REF" ]; then echo "##[set-output name=branch_name;]$(echo ${GITHUB_REF#refs/*/})"; else echo "##[set-output name=branch_name;]$(echo ${GITHUB_HEAD_REF})"; fi
```
ca-scribner, line 54: I think this'll be easier to read as a multi-line script, and I believe it still works.  So instead of the one-liner here you could have:
```
    if [ -z "$GITHUB_HEAD_REF" ]; then
      echo "..."
    else
      ...
    fi
```

base-notebook/Dockerfile
```diff
@@ -1,24 +1,27 @@
...
+ARG BASE_CONTAINER=master
+FROM $BASE_CONTAINER
 USER root
 ENV PATH="/home/jovyan/.local/bin:${PATH}"
 
 RUN pip --no-cache-dir install --quiet \
       'kfp==1.0.0' \
       'kfp-server-api==1.0.0' \
-      'kfp-tekton==0.3.0' \
+      'kfp-tekton==0.2.0' \
```
ca-scribner, line 9: bump this version back up to 0.3.0.  For any version differences, we likely take the newer one between master and this branch

base-notebook/Dockerfile
```diff
@@ -1,24 +1,27 @@
...
       'kubeflow-fairing==1.0.1' \
-      'ml-metadata==0.24.0' \
```
ca-scribner, line 11: I think we want to have this addition.  It was added after I branched off master

base-notebook/Dockerfile
ca-scribner, line 9: Bump this up to 0.3.0.  In general if there are version differences, we probably want the newer one

ca-scribner:
closes #122

ca-scribner:
@zachomedia this is ready for you to review.  The main change is that instead of:
* dev running gpu-jupyter to create a Dockerfile
* dev copies above Dockerfile into the top of `base-notebook-gpu`
this update does:
* CI runs gpu-jupyter to make a `Dockerfile` that is a gpu-enabled equivalent to the docker-stacks file we inherrit
* `base-notebook` for cpu and gpu can be the same (because all gpu-specific stuff is in the preceeding image) and now takes an ARG for its base container (used by CI to build either CPU or GPU flavor

That, and there's now CI enabled for PR's to master in addition to merging into master, which allows debugging CI without actually merging into master.

ca-scribner (approved):

sylus:
@zachomedia can you review this today?

Issue #146, opened 2020-11-03
Split kubeflow-containers PRs to smaller chunks for review


ca-scribner:
Split #107 and #122 into smaller, easier to review chunks.  Prioritize splitting the parts that need cloud team review from the others so that Will/Zach have less to review.

ca-scribner:
PR #129 is being split into pieces that implement the following improvements:
* **NOTE**  Start from current `master` - #129 started from an old commit which has many older versions for many packages. 
 Make sure all pip/conda installs have the `fix-permissions` fixes in place (see #129's readme.md for description of this problem)
* [ ] Refactor base-notebook-gpu, instead pulling anything gpu-specific out of it and creating an `upstream-equivalent-notebook-gpu`.  `upstream-equivalent-notebook-gpu` will be built on the fly by this CI from a SHA-pinned pull of [GPU-Jupyter](https://github.com/iot-salzburg/gpu-jupyter) to create the GPU equivalent to [docker-stacks' datascience-notebook](https://github.com/jupyter/docker-stacks), which is our CPU upstream image.  This lets both CPU and GPU versions of `base-notebook` be identical except for the upstream image
* [ ] Update all Dockerfiles to take the image specified in `FROM IMAGE` as a build arg.  Update CI to use this.  Also use this feature to pass SHA-pinned versions of the built images rather than generic `:master` images
* [ ] Encapsulate the repetitive steps in CI (the build instructions for each step, which are repeated for each) into a helper script to make it easier to read the CI and easier for dev's to test building locally
* [ ] Add build caching to CI (see CI helper script from #129, which pulls the latest version of the image before trying to build so it can benefit from cached layers).  #129 had a bug where it would not tag things with a branch name - need to address this
* [ ] Combine the CPU and GPU Dockerfiles and refactor the directory structure to remove CPU/GPU subdirs.  Update CI to use this structure
* [ ] Add `build.sh` files for CPU and GPU builds, as well as a `build_all.sh` file at the high level, to help future devs (these can be pulled from the original PR)
* [ ] Include the updated readme.md's from entire repo and each notebook to help future devs

PR #156, opened 2020-11-23
feat: reuse cached layers to reduce build times

JessicaBarh:
Add layer caching to the CI to speed up deployment/testing. CI now does docker pull this_image:master prior to each docker build in order to pull the most recently built image in an attempt to get layers that docker build can cache from. If a perfect cache hit, each image builds in <5 min instead of 15-35 min. This also prevents rebuilding images unnecessarily (eg: if making a CPU-only change, the GPU images will just pull from cache rather than rebuild). In the worst case (editing base-notebook) this adds ~3-5min to build.

This Relates to PR [#129](https://github.com/StatCan/kubeflow-containers/pull/129) that is being split up into smaller PRs for review - for issue [#146](https://github.com/StatCan/kubeflow-containers/issues/146)


.github/workflows/build-cpu.yml
```diff
@@ -18,7 +18,7 @@ on:
   pull_request:
     branches: 
       - master 
-      - combine-all-cpu-gpu-dockerfiles #DEBUG
+      - encapsulate-repetitive-steps-in-CI #DEBUG
```
ca-scribner, line 21: This can be replaced by what you have in an earlier PR
```
  pull_request:
    branches: 
      - master
```
(plus handling the two different naming conventions for branches based on if they're a PR or merge)

If you had that code, you wouldn't need to add a branch to the `pull_request` CI each time you want to test a new branch, it would just handle that automatically

scripts/build_push.sh
```diff
@@ -4,15 +4,17 @@ set -e
```
ca-scribner, line 4: The build_push.sh could be more general if we wanted.  It could just take lists of
```
pull_before_build = [...]  # (for caching)
tag_output = [...] # (for applying arbitrary number of tags that then get pushed)
```
I was lazy before and didn't make lists, but might make it more reusable.

ca-scribner (approved):
lgtm but should be reviewed by Blair or Brendan

PR #157, opened 2020-11-23
Fix tf cudnn, buggy tf install in gpu image

ca-scribner:
Bumped base image in GPU stack to use NVIDIA's CUDNN base, enabling GPU usage

Fixed bug where tensorflow is accidentally installed in `base-notebook-gpu` by installing `kubeflow-metadata`, which causes a later bug in machine-learning-notebook.  Also bumped tensorflow version in GPU to match that in CPU

### DIFF
--- a/.github/workflows/build-gpu.yml
+++ b/.github/workflows/build-gpu.yml
@@ -15,6 +15,9 @@ on:
   push:
     branches:
       - master
+  pull-request:
+    branches:
+      - master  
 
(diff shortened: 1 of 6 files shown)


ca-scribner:
will squash on merge.  

summary: 
* rebases gpu image on cudnn version
* fixes a bug in installing tensorflow
* removes Azure CLI SHA checking because it seems more frustrating than useful, but we can leave that in

PR #159, opened 2020-11-25
Draft: Mixin approach

blairdrummond:
I split into 2 commits, one which deletes everything and one which adds everything.

Still a work in-progress (need to pin things) but all the images are [on dockerhub](https://hub.docker.com/u/blairdrummond):

- The total build time is less than 40 minutes for everything! And that's *not* with any layer caching.

- There is added functionality here as well, such as a spark base.

- No need to rebuild docker-stacks with a GPU base, as we put a GPU layer on-top of docker-stacks. (It's tested and works great


ca-scribner:
one thing I've seen some other dockerfile-orchestrating repos do that is nice is put the output into a `.output` or `.build` folder to make it a bit clearer that that content is more a product than something people should directly change.  They also had a readme in the `.build` folder stating the files were auto generated, and ideally a header in each the `Dockerfile`s saying they were autogenerated).

blairdrummond:
Closing in favour of #160

Issue #150, opened 2020-11-05
Need to run ```source .bashrc``` to get conda commands to work in new terminal


Collinbrown95:
Description
-----------
I was setting up a python virtual environment with conda in a jupyterlab notebook server (minimal-notebook-cpu image) and when I tried to activate a conda environment using ```conda activate <my-environment-name>``` I got:

```bash
CommandNotFoundError: Your shell has not been properly configured to use 'conda activate'.
To initialize your shell, run
    $ conda init <SHELL_NAME>
```
So I ran ```conda init bash``` as per the instructions and ran ```source .bashrc```, when I do this ```conda activate <my-environment-name>``` works as expected.

However, if I open a new terminal I need to run ```source .bashrc``` again to allow me to activate my virtual environment in the new terminal.

Expected Behaviour
---------------
It would be convenient if conda commands worked by default when you open a new terminal so that you don't need to run ```source .bashrc``` every time.

Issue #147, opened 2020-11-04
Look into a more maintainable model for these images


blairdrummond:
The current maintenance of these images is a pain. Resolving this will require more than one ticket, but the purpose of this issue is to try to investigate some ideas.

## Integration with docker-stacks

How can we keep our base images up-to-date better, without accidentally introducing breakage.

Thoughts:
- We need to stay better up-to-date with docker-stacks
- We need integration testing for checking compatibility with our JupyterLab plugins
- We probably need better automation for monitoring upstream

## Our base | ML-Tooling could be split up better

Our ML-Tooling approach probably needs work. Some things like conda environments (possibly pre-packaged ones) might be useful. In general, the ML-Tooling is up to the user, so maybe we should be less involved in that space.

# General ideas

## Buildah?
The makefile based approach for creating the Dockerfiles is interesting. Another tool that might be worth looking at is [buildah](https://github.com/containers/buildah).

## Process for "one-off" user images?

We might also consider a better workflow for accommodating user requests within an image. As an example, some members of my team recently needed access to Selenium & Chrome in order to write some web-scraping code. They had a DAaaS team-member build one for them manually so that they could use this in a Jupyter Notebook. This approach seems too cumbersome. Ideally users would have a simple workflow for creating their own images.

## Ansible &/or a manual access into the images?

Weird idea, could we simply open a door into the images via ansible? Instead of rebuilding the images, could we open up the means to do ad-hoc installs into the container at run-time instead of build time? (This may be **WORSE** since it bypasses scanning.)

blairdrummond:
CC @justbert @ca-scribner @brendangadd

ca-scribner:
Generally agreed - good summary!

# Maintenance/docker-stacks

* I don't know if staying up to date with docker-stacks helps or hurts us from maintainability (more frequent syncs leads to more frequent breaks), but it would be good to keep current if it doesn't add effort
* upstream monitoring (both docker-stacks and any other pinned dependency) would be nice.  It would be neat if we could have a declarative way of keeping our dockerfiles up to date for packages we know can be trusted.  Not sure which way is best, but we'd like to achieve something like:
```
FROM docker-stacks/minimal-notebook:KEEP_UPDATED
RUN apt get install some_program=KEEP_UPDATED

# Or maybe: 
RUN SOME_PROGRAM_KEEP_UPDATED_PLACEHOLDER

# Something that shouldn't be updated
RUN apt get install some_old_program=1.2.3
```
[...]

Issue #137, opened 2020-10-16
Consider using zsh + oh-my-zsh as default shell?


blairdrummond:
Might make the out-of-the-box scripting experience a bit nicer.

blairdrummond:
I think zsh is much nicer than bash, it's now packaged as the default in MacOS (which is probably more prevalent in users than linux, whether I like it or not), and oh-my-zsh provides nice git support. Additionally, I do not see many drawbacks.

CC @brendangadd

sylus:
Why are we removing this then? Zshell is what most developers are familiar with including myself, and assume most data scientists?

Can we keep this as is?

sylus:
@zachomedia

brendangadd:
To clarify the situation:
* The change in default shell came in as part of an unrelated architectural change – I missed it in review; that's my bad
* Changes like this need proactive discussion, so it's being reverted
* The PR that reverts this behaviour (#169) does not close this issue, which can remain open for (proactive) discussion
* `zsh` remains installed in the image, so users who want to use it still can

Actions I've taken:
* Moved #169 to Review status (LGTM. Can you take a look @ca-scribner?)
* Moved this issue out of this sprint as it is not a priority
* Asked @Colette-G to have UX gather some evidence about this so that objective discussion is possible

/cc @sylus @blairdrummond

bryanpaget:
Or take it a step further with [Starship](https://starship.rs/).

brendangadd:
@bryanpaget Starship looks cool, and compatible with Bash (default) and zsh (included in image). Maybe worth a separate issue if you feel it's a major UX boost.

Issue #121, opened 2020-09-17
Tensorflow does not work in machine-learning-notebook (and others?)


ca-scribner:
Running something like 
```
import tensorflow as tf
tf.test.is_gpu_available(
    cuda_only=False, min_cuda_compute_capability=None
)
```
returns the message `AttributeError: module 'tensorflow' has no attribute 'test'`.  This is repeated for anything `tf.*`.  Similar issues pop up if using `kubeflow metadata`/`ml medatada`, likely for the same reason.

Can be fixed by 
```
pip uninstall tensorflow  # Note this is 2.1.0 as of writing
pip uninstall tensorflow-estimator
pip install tensorflow==2.3.0
```
Unsure if this fix works because we're installing a different version, or if the 2.1.0 install is corrupt in general?  Tensorflow gets installed by accident in `base-notebook` when installing `kubeflow-metadata`, so that might be related?

saffaalvi:
This may have been fixed when the Tensorflow package was updated to version 2.2.0:
 https://github.com/StatCan/kubeflow-containers/blob/423b70c25f67e66698f54d0d849652ddd280e795/machine-learning-notebook/cpu/Dockerfile#L7

Now, when I run the command you gave as an example, there is no more error but instead just a deprecation warning (only shows up the first time it's ran):
```
WARNING:tensorflow:From <ipython-input-1-a32398aaaac2>:2: is_gpu_available (from tensorflow.python.framework.test_util) is deprecated and will be removed in a future version.
Instructions for updating:
Use `tf.config.list_physical_devices('GPU')` instead.
False
```
![Screen Shot 2020-10-19 at 12 46 00 PM](https://user-images.githubusercontent.com/57377830/96481176-2fbf3d80-1209-11eb-89ca-8071884c5a09.png)

Tensorflow is also no longer installed in `base-notebook ` (also a result of version update), so the only notebook that has it installed is `machine-learning-notebook/cpu`.

saffaalvi:
Tested out [kubeflow-metadata demo](https://github.com/StatCan/jupyter-notebooks/blob/2b40391ec0b4a8fc795baedc124f572f6b97a0c7/kubeflow-metadata/overview.ipynb)  on custom image: machine-learning-notebook-cpu:129ee3e50de291c4fe65397a614f5cccb379d5fd with the latest version updates. 

Everything seems to be working well so this issue will be resolved when latest image commits are updated in kubeflow-manifest.

[...]

PR #160, opened 2020-11-26
Refactor and streamline builds using Matrix Strategy

sylus:
In general I and @zachomedia are a fan of this approach, and think it hits a few of the pain points:

a) All are based on a single template
b) The final output is still exported and stored in Git
c) All of the builds are done in a Matrix Strategy
d) The CI run finishes in about 40 minutes

> Thanks @blairdrummond for working on this and proposing this P.R.

Screen Shot 2020-11-26 at 3 03 06 PM


sylus:
@blairdrummond is the CUDA bug fixed here for the GPU images as well?

blairdrummond:
@sylus I'd appreciate @ca-scribner 's review on the GPU bugs to see, but I think yes:

Screenshot from 2020-11-15 16-33-39

Screenshot from 2020-11-15 11-05-23

sylus:
@brendangadd is there additional information and / or question you require for this for us to accept this approach?

ca-scribner:
migrating some comments over from #159:

@sylus @zachomedia when Blair walked me through this and I was torn. I really liked how it made it easy to identify and maintain the install instructions for various mixins (eg: Installing jupyterlab? go to docker-bits/6_JupyterLab.Dockerfile), but it also felt like it side steps some benefits of Docker (ex: mixins lead to many more layers because we aren't combining dissimilar installs into the same RUN, layer caching gets less transparent, ...).

I also worried it made it harder for someone new to join in and contribute (need to learn the makefile structure, etc). But counter to that, maintenance of each individual piece because far easier and more transparent, so maybe it is a net positive.

Am I worried about nothing, or do you have suggestions to work around this?

Suggestion for **Blair**: one thing I've seen some other dockerfile-orchestrating repos do that is nice is put the output into a .output or .build folder to make it a bit clearer that that content is more a product than something people should directly change. They also had a readme in the .build folder stating the files were auto generated, and ideally a header in each the Dockerfiles saying they were autogenerated).

ca-scribner:
@sylus this does install the missing CUDNN drivers.  I **think** it works, but I'm not sure.  The CUDNN drivers get installed after a lot of other stuff instead of at the top in the NVIDIA-provided base image, and I don't know if that's a problem.  Do you?  If installing the NVIDIA drivers that late is ok, why does [gpu-jupyter](https://github.com/iot-salzburg/gpu-jupyter) exist at all?

ca-scribner:
If we decide to go with this, I think there's still some refactoring to discuss.  The r-studio image now has spark, the CUDA driver version is bumped, and I think there's some other stuff add/subtracted, so someone should review with software and versions in mind.

ca-scribner:
I think I like this approach better if we refactor to be a full mixin approach rather than a hybrid mixin/dag.  The `docker-bits` shouldn't assume the order/mutual-exclusivity of how `Dockerfile` snippets go together.  Instead it should be a repo of `Dockerfile` snippets for each software/version (and maybe a `latest` that points to the latest snippet for each), so something like:
- nvidia drivers
  - 10.0
    - Dockerfile.snippet
    - ..any other supporting files that should be in the build context
  - 11.0
    - Dockerfile.snippet
    - ...
  - latest
    - points to 11.0 or whatever we want `latest` to build from, giving us an easy way to update all images that we want to always be up to date

Then separately (but maybe in same repo) we can have a collection of template files that represent the different end products (`minimal-notebook-cpu`, `machine-learning-gpu`, ...).  Each template definition can include an ordered list of the `snippets` that form their output, then the build makefile can be in charge of executing the build of all templates.  That way, just like we've encapsulated the definition of each software's Docker definition into easy to find pieces, we do the same to the DAG that defines the actual `Dockerfile`.  So something like this:
- templates
  - minimal-notebook-cpu
    - definitionFile
- templates
  - minimal-notebook-gpu
    - definitionFile
- ...

where `minimal-notebook-gpu`'s definitionFile could be:
```
nvidia_drivers/11.0
kubeflow/latest
tensorflow/2.3.0
...
```

...ideally I'd actually like to be able to refer to a past definition file in there too, that way we could benefit from intermediate Dockerfiles in addition to the nice and maintainable snippets.  Something like:
```
templates/minimal-notebook-gpu
someAdditionalSoftware/latest
...
```

blairdrummond:
@ca-scribner 

## Layers

- All layers shared within the docker-bits will be shared in the ACR, if possible. E.g VSCode-Spark and JupyterLab-Spark will share everything up to the UI. 

- The layers *are* built on many machine in parallel at once (which is redundant, true), but that's what lets the failure-free parallel CI process work, and the images are built faster as a result.
 
- Also, the compute time is free on public github actions (plus, it only takes 40 minutes to build **everything** anyway).

## GPU
- I have screenshots there of PyTorch and Tensorflow using the GPU... If you have another test-case to run then we can check that?
- The NVidia base images are super simple when you dig into them. I read through their image hierarchy. [repo](https://gitlab.com/nvidia/container-images/cuda/-/blob/master/dist/11.1/ubuntu18.04-x86_64/runtime/cudnn8/Dockerfile)
- I really don't think we're missing anything there...

## On templates
I have to hard-disagree on the template-per-target thing. 

- I think having a single template engine is much simpler, and means that changes only need to be made "locally". One of the aims of this design was to minimize duplication.
- Inheritance and template-per-target both struggle with multiplication. The DAG in docker-bits is really letting you "multiply" features, without multiplying the amount of maintained boiler-plate code.

## DAGs are our friend

- Also, the sequence of application is important. It's good to be clear what order things get applied in.
- Some steps are mutually exclusive, so the convention that you only take one `docker-bit` from any layer of the DAG seems nice.

## Indicate that the output files are generated
Sure thing. I'm not gonna muck around in @sylus 's PR, but we can add more documentation. Sounds like we could add that to the Makefile output target.

```Makefile
.output:
    mkdir -p $(OUT)/ $(TMP)/
    echo "\# These files are automatically generated" > $(OUT)/README.md
    echo 'Treat these as "read only" and modify the `docker-bits` instead. Then compile with `make all`'  >> $(OUT)/README.md
```

# Feature Related Changes

I 100% agree, there are feature changes that deserve discussion. For example:

- I added conda-environment notebooks.
- The visuals of Conda environments in Notebooks could be nicer. But I think this is relatively minor.
- Yes, Spark is in the R image, I added it for Spark-R.
- I made zsh (actually oh-my-zsh) the default shell, and they start in a conda environment with minio auto-loaded.
- I added [Pluto.jl](https://github.com/fonsp/Pluto.jl) for Julia notebooks, which was requested in the AAW slack channel by a user.
- I added VS-Code, of course.
- I added [PGWeb](http://sosedoff.github.io/pgweb/) since Postgres for geospatial is on the horizon (and also, a client doesn't hurt anyone.)
- I removed the azcli pin.
- :warning: I unpinned the JupyterLab extensions. They were (predictably) breaking, so I built without pins, but can test the components and repin with the new versions.

# Conclusion

I can pull the new features out and add them as a separate PR for review, but I stand by the structure here.

blairdrummond:
Resolves #150 , #147 , #137 , #136 , #122 , #121 , #107  , #27 , #48 ,

zachomedia (approved):
This is LGTM. Let's merge and go from there!

Issue #149, opened 2020-11-02
Add i18n to RStudio


blairdrummond:
Look into Bilungualism options for RStudio

wg102:
The i18n of Rstudio can only be partial. It seems the menus/text are still in English, but the inside (the text, and the error messages) are able to be in French, if one changes the Environment variable to French

This is the code to add the French Locale and set it.
```
RUN echo "fr_CA.UTF-8 UTF-8" > /etc/locale.gen && \
    locale-gen
# Configure environment
ENV CONDA_DIR=/opt/conda \
    LC_ALL=fr_CA.UTF-8 \
    LANG=fr_CA.UTF-8 \
    LANGUAGE=fr_CA.UTF-8
```

The next step is to find how to detect the language, and in runtime set the environment variable.
Current idea is to create a 'transparent layer' similar to remote-desktop dashboard but without UI that will check the browser language before it opens to the user.

wg102:
To have R-studio in ‘French’ it needs `LANG `to be setup with the correct locale.  
The way it was decided is to “take the active language in the KF UI and automatically submit it as part of the "New Server" payload, and make the controller pass that locale as an env var (as you suggest) to all notebooks it launches. Then any container can find locale information at a known location and do with it as it pleases.” Which means to send it when creating a new notebook. 
The equivalent of testing with `docker run -e LANG=fr_CA.UTF-8 imageTag`, which overrides whatever value of that environment variable set in the docker file.
The changes therefore need to apply to multiple places, and some things need to be verified
Kubeflow-container to add the French locale (**TODO**: decide where to add locale: in the base file, or in r-studio file). 
Jupyter-api to add the language detection (with a controllable UI). Whatever the setting in those will inject: 

- LANG
- LANGUAGE
- LC_ALL ?

R-Studio (image) only needs LANG. 
The R-studio in remote desktop needs both LANG and LANGUAGE.
Other applications might be impacted when changing the locales. (to be investigated)

wg102:
From what I gathered, the way to have environment variables would be through the PodDefault (see https://www.kubeflow.org/docs/notebooks/setup/ step 12)

wg102:
The short answer for this issue, is to have the Environment variable LANG set to the wanted language. For this to work, the locale for that language needs to also be available (ex: fr_CA.UTF-8).

wg102:
This issue is split in two part, 

[...]

PR #164, opened 2020-12-01
feat(rstudio): set LANG based on KF_LANG

wg102:
Related to https://github.com/StatCan/kubeflow-containers/issues/149

KF_LANG is planned to be included as an environment variable in the pod when creating the notebook.


blairdrummond:
@wg102 , the `output/` folder is generated from the Makefile. You actually want to put this change here

https://github.com/StatCan/kubeflow-containers/blob/master/resources/start-custom.sh

sylus:
Resolves #149 I made the fix and pushed it up

PR #166, opened 2020-12-02
Clean up the images

blairdrummond:
- Remove Spark (was causing issues)
- Remove pgweb & Pluto.jl
- Merge Jupyterlab and VSCode images

Small edits: also fixed a line in the Makefile and added torchtext in the default PyTorch.

End result: We're down to only 4 images :-D

We will re-add spark, but more gracefully. I don't like the `all-spark-notebook` from Dockerstacks.


sylus:
Why are we merging vscode and jupyter? Isn't it nice they separate?

@blairdrummond

blairdrummond:
Yep last one. Was based on feedback from @brendangadd , main reason is to simplify the user flow. Too many images being overwhelming.

The Spark thing was my bad. It was causing issues in R-Studio because it messes with the R installation, but we'll add it into the CPU directly.

PR #169, opened 2020-12-02
feat(zsh): remove zsh as default shell

blairdrummond:
This removes `zsh` as the default shell, which reverts to the previous behaviour.

I have an old issue here (#137), suggesting that we add it.


ca-scribner (approved):
lgtm

ca-scribner:
I believe the CI will resolve itself because of a separate patch into master

PR #193, opened 2021-01-18
feat(minio): Automate setup MinIO credentials

### LINKED ISSUE #189: Automate setup/refresh of MinIO credentials
MinIO credentials are mounted to notebook/pipeline pods in `/vault/secrets`.  These credentials have lifespans (1 day?  1 week? 
 can't remember) and are refreshed periodically for pods as they're running.  

Current recommendations for using MinIO are to `source /vault/secrets/...` to set up a single MinIO setup with the current MinIO credentials whenever users need them.  This causes issue because:
* It is a chore that is a barrier to entry for using MinIO (users must know how to set up MinIO, extra tasks for every use, etc)
* When credentials become stale this must be redone manually

# Proposal: 

Somehow automate the setup/refreshing of MinIO credentials so that users do not need to do this themselves.  Some possible ideas:
* A background process running on all notebook servers/pipeline components that watches for changes to `/vault/secrets/...` and  communicates them to MinIO (either directly into the `~/.mc/config.json` file, or maybe using MinIO interface `mc config host add ...`?)
* Configure MinIO to use a `config.json` file that Vault updates directly, negating the need for a watcher process
* ...?

Nancy-Badran:
Resolves #189


resources/check-tenant.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# This script checks if the tenant if newer than a little "lockfile" .
+# Every time that the user calls mc, the for loop checks to see if any vault secrets are newer than the most recent lockfile. 
+#If the secret is newer, then it reinitializes the client right away.
+for tenant in standard premium; do
+    if [ ! -f /tmp/.minio-$tenant ] || [ /vault/secrets/minio-$tenant-tenant-1 -nt /tmp/.minio-tenant ]; then
```
blairdrummond, line 6: This was my bad I think, but there is a missing `$` in the second `.minio-tenant`.

blairdrummond (changes requested):

resources/check-tenant.sh
Nancy-Badran: can you plz confirm what I did in https://github.com/StatCan/kubeflow-containers/blob/ee176847b2191bfcdb2de198de06422b93636796/docker-bits/4_CLI.Dockerfile#L46,  to avoid the permission denied on /usr/local/bin/mc

output/JupyterLab-CPU/Dockerfile
```diff
@@ -88,7 +88,7 @@ RUN curl -LO "${KUBECTL_URL}" \
     wget --quiet -O mc "${MC_URL}" \
     && echo "${MC_SHA} mc" | sha256sum -c - \
     && chmod +x mc \
-    && mv mc /usr/local/bin/mc \
+    && mv mc /usr/local/bin/mc-wrapped \
```
blairdrummond, line 91: Yep, this looks right

resources/start-custom.sh
```diff
@@ -31,6 +31,8 @@ if [ -n "${KF_LANG}" ]; then
     fi
 fi
 
+# copy check-tenant.sh to /usr/local/bin/mc
+cp /check-tenant.sh /usr/local/bin/mc
```
blairdrummond, line 35: Have you tested this? I don't think this will work... start-custom.sh does not run as root, so I would not expect this command to have permissions to do this.

blairdrummond:
Also, you should delete this from the existing code:

https://github.com/StatCan/kubeflow-containers/blob/ee176847b2191bfcdb2de198de06422b93636796/resources/helpers.sh#L4-L11

Your PR basically replaces that functionality

resources/start-custom.sh
blairdrummond: I think you'll need to use a COPY in the Docker files

Nancy-Badran: yeah , it give me a permission denied , so I used COPY in Dockerfile and change the mode of access to /usr/local/bin/mc, but I didnt delete it from start-custom.sh . Ok So I will delete the` cp /check-tenant.sh /usr/local/bin/mc` from the start bash.

docker-bits/4_CLI.Dockerfile
```diff
@@ -43,7 +43,7 @@ RUN curl -LO "${KUBECTL_URL}" \
     wget --quiet -O mc "${MC_URL}" \
     && echo "${MC_SHA} mc" | sha256sum -c - \
     && chmod +x mc \
-    && mv mc /usr/local/bin/mc \
+    && mv mc /usr/local/bin/mc-wrapped \
```
ca-scribner, line 46: I think we could be clearer with the name here.  `mc-wrapped` suggests it is actually wrapped already, whereas we are renaming it so we can wrap it later.  Maybe `mc-original` or something like that?

output/RStudio/check-tenant.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
ca-scribner, line 1: Should we rename this to be clearer of the purpose, and be clearer about purpose in the top line of the script? `check-tenant.sh` doesn't to me say "I am a replacement wrapper for `mc` that will update minio tenants if they are stale before running". I think someone glancing at this file quickly wont realize you should run your `mc` commands through it, unless maybe they catch the call to `mc-wrapped` on the last line (and know what `mc-wrapped` is)

Maybe call it `mc-tenant-wrapper.sh` or something like that? And then a good first line comment in the file that says exactly how this is used.

output/RStudio/check-tenant.sh
```diff
@@ -0,0 +1,13 @@
...
+# This script checks if the tenant if newer than a little "lockfile" .
+# Every time that the user calls mc, the for loop checks to see if any vault secrets are newer than the most recent lockfile. 
+#If the secret is newer, then it reinitializes the client right away.
+for tenant in standard premium; do
+    if [ ! -f /tmp/.minio-$tenant ] || [ /vault/secrets/minio-$tenant-tenant-1 -nt /tmp/.minio-$tenant ]; then
```
ca-scribner, line 6: How does this handle `minio-$tenant-tenant-1` files that are missing? I can think of three cases

1. minio-standard-tenant-1 exists, minio-premium-tenant-1 does not exist
2. minio-standard-tenant-1 does not exist, minio-premium-tenant-1 exists
3. minio-standard-tenant-1 does not exist, minio-premium-tenant-1 does not exist

I'm not 100% sure what will happen in these scenarios, but my gut says at least one will break. The behaviour I think we want is that any missing/corrupt file will be skipped without affecting the others.  I could see either if the first file is missing that the second doesn't get updated, or that if the second is missing that it will actually get overwritten by the first

output/RStudio/check-tenant.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# This script checks if the tenant if newer than a little "lockfile" .
+# Every time that the user calls mc, the for loop checks to see if any vault secrets are newer than the most recent lockfile. 
+#If the secret is newer, then it reinitializes the client right away.
+for tenant in standard premium; do
```
ca-scribner, line 5: Should we explicitly list these tenants here, or should we loop over files matching a pattern? Something like loop over all `/vault/secrets/minio-*`. That way we don't have to modify this later

output/RStudio/check-tenant.sh
```diff
@@ -0,0 +1,13 @@
...
+    if [ ! -f /tmp/.minio-$tenant ] || [ /vault/secrets/minio-$tenant-tenant-1 -nt /tmp/.minio-$tenant ]; then
+        source /vault/secrets/minio-$tenant-tenant-1
+        mc config host add $tenant $MINIO_URL $MINIO_ACCESS_KEY $MINIO_SECRET_KEY
```
ca-scribner, line 8: We should name the `$tenant` here the same as they are named in the mounted directories under `~/minio/*`. Those directories right now are given the full name, eg: `standard-tenant-1`.

ca-scribner (changes requested):
Note: I accidentally wrote comments on `output/RStudio/check-tenant.sh` but they're supposed to be on `resources/check-tenant.sh`.  Read it as if I meant to edit `resources/check-tenant.sh` please :)

output/RStudio/check-tenant.sh
Nancy-Badran: yeah I will do  it to loop over `/vault/secrets/minio-*`

output/RStudio/check-tenant.sh
Nancy-Badran: In /vault/secrets/minio-* it is already named as same as they are named in mounted directories , I will rename it also in /tm/*

ca-scribner: yeah I think a nice benefit of looping over the files in the dir is you'll get the same full names as what are mounted in `~/minio`

output/RStudio/check-tenant.sh
ca-scribner: I'm sure you'll catch it on implementation, but all the files have two copies (one using environment variables, one written in JSON), so you'll have to filter to only handle one of the two

Makefile
```diff
@@ -73,7 +73,7 @@ CPU: .output
 # Only one output version
 RStudio: CPU
 	mkdir -p $(OUT)/$@
-	cp -r resources/* $(OUT)/$@
+	cp -r  resources/* $(OUT)/$@
```
ca-scribner, line 76: tiny thing - there's a double space here before resources instead of single

docker-bits/4_CLI.Dockerfile
```diff
@@ -43,14 +43,11 @@ RUN curl -LO "${KUBECTL_URL}" \
...
     curl -sLO https://aka.ms/InstallAzureCLIDeb \
     && bash InstallAzureCLIDeb \
     && rm InstallAzureCLIDeb \
   && \
     wget -q "${OH_MY_ZSH_URL}" -O /tmp/oh-my-zsh-install.sh \
-    && echo "${OH_MY_ZSH_SHA} /tmp/oh-my-zsh-install.sh" | sha256sum -c
-
-# Have to use the zsh-installer in the entrypoint!
-COPY helpers.sh /tmp/helpers.sh
+    && echo "${OH_MY_ZSH_SHA} /tmp/oh-my-zsh-install.sh" | sha256sum -c
```
ca-scribner, line 53: was this an intentional change to remove the helpers.sh?  Just seems unconnected so asking to make sure.  If this was intentional, sounds good to me

output/RStudio/mc-tenant-wrapper.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+# This script checks if the tenant if newer than an exiting one.
+# Every time that the user calls mc, the for loop checks to see if any vault secrets are newer than the most recent lockfile. 
+# If the secret is newer, then it reinitializes the client right away.
+
+for  f in $(ls /vault/secrets/minio-* | grep -v json); do
```
ca-scribner, line 6: We can make this more robust.  We want to find files with no file extension.  This filters out files ending in `.json`, but what if there were others?  If we add a `.yaml` file here for some other application, we'll break this wrapper unexpectedly.  I think you can more directly find the files you're looking for.

One idea is to `grep -v -E '\..*'`, which I think would match anything that does not have have an extension.  Another is to use `find . -type f ! -name "*.*"` which finds only files without extensions

output/RStudio/mc-tenant-wrapper.sh
```diff
@@ -0,0 +1,14 @@
...
+ tenant=$(basename "$f" | sed 's/.json$//' | sed 's/^minio-//') # remove the json suffix and minio- prefix
```
ca-scribner, line 7: With a finer search for the for loop above, the `sed` for json can be removed here

output/RStudio/mc-tenant-wrapper.sh
```diff
@@ -0,0 +1,14 @@
...
+# This script checks if the tenant if newer than an exiting one.
+# Every time that the user calls mc, the for loop checks to see if any vault secrets are newer than the most recent lockfile. 
+# If the secret is newer, then it reinitializes the client right away.
+
+for  f in $(ls /vault/secrets/minio-* | grep -v json); do
+ tenant=$(basename "$f" | sed 's/.json$//' | sed 's/^minio-//') # remove the json suffix and minio- prefix 
+ if [ ! -f /tmp/.minio-$tenant ] || [ $f -nt /tmp/.minio-$tenant ]; then
+     source $f
+     mc config host add $tenant $MINIO_URL $MINIO_ACCESS_KEY $MINIO_SECRET_KEY
+     touch /tmp/.minio-$tenant
```
ca-scribner, line 11: this should really be in a subshell, I think like this:

```
(
  source $f
  mc config host add $tenant $MINIO_URL $MINIO_ACCESS_KEY $MINIO_SECRET_KEY
)
```

Since we're using environment variables to pass info from `source $f` to the `mc` command, we should use a subshell to avoid accidentally putting those variables in the current shell.  This wont matter most of the time, but if something goes wrong I think it could

output/RStudio/mc-tenant-wrapper.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+# This script checks if the tenant if newer than an exiting one.
+# Every time that the user calls mc, the for loop checks to see if any vault secrets are newer than the most recent lockfile. 
+# If the secret is newer, then it reinitializes the client right away.
+
+for  f in $(ls /vault/secrets/minio-* | grep -v json); do
+ tenant=$(basename "$f" | sed 's/.json$//' | sed 's/^minio-//') # remove the json suffix and minio- prefix 
+ if [ ! -f /tmp/.minio-$tenant ] || [ $f -nt /tmp/.minio-$tenant ]; then
+     source $f
+     mc config host add $tenant $MINIO_URL $MINIO_ACCESS_KEY $MINIO_SECRET_KEY
```
ca-scribner, line 10: have you tried running this script, which calls `mc config ...`, in a built container?  I think this is going to make an infinite recursive loop, because `mc` is this script.  I believe you want to call `/usr/local/bin/mc-wrapped`?

output/RStudio/mc-tenant-wrapper.sh
```diff
@@ -0,0 +1,14 @@
...
+     touch /tmp/.minio-$tenant
+ fi
+done
+/usr/local/bin/mc-wrapped $@
```
ca-scribner, line 14: Put `/usr/local/bin/mc-wrapped` into a variable at the top, that way it is really clear.  Something like `mc_executable=/usr/local/bin/mc-wrapped`, then you can later invoke that here (and if I'm right about the recursion problem above, there as well)

resources/start-custom.sh
```diff
@@ -17,8 +17,8 @@ if [ ! -f /home/$NB_USER/.zsh-installed ]; then
       conda init zsh
     fi
 
-    cat /tmp/helpers.sh >> /home/$NB_USER/.bashrc
-    cat /tmp/helpers.sh >> /home/$NB_USER/.zshrc
+    #cat /tmp/helpers.sh >> /home/$NB_USER/.bashrc
+    #cat /tmp/helpers.sh >> /home/$NB_USER/.zshrc
```
ca-scribner, line 21: same question as above for helpers.  Intentional change?

ca-scribner (changes requested):

docker-bits/4_CLI.Dockerfile
brendangadd: Please preserve trailing newlines.

docker-bits/∞_CMD.Dockerfile
```diff
@@ -3,6 +3,8 @@
 WORKDIR /home/$NB_USER
 EXPOSE 8888
 COPY start-custom.sh /usr/local/bin/
+COPY mc-tenant-wrapper.sh /usr/local/bin/mc 
+RUN chmod a+rx /usr/local/bin/mc
```
brendangadd, line 7: Does `COPY` preserve the file mode? Can we set the correct mode on the file in the repo and avoid the `chmod`? /cc @blairdrummond

Nancy-Badran: I think yeah, COPY preserve the file mode so we can avoid to change mode here

blairdrummond: Yep COPY should preserve file mode

resources/start-custom.sh
blairdrummond: This is intentional I think, but should be removed not just commented out

output/RStudio/mc-tenant-wrapper.sh
blairdrummond: Actually I think this would be fine, unless you called `source /usr/local/bin/mc`, if I'm not mistaken. I think exports run within this script wouldn't affect the calling shell

blairdrummond: Test:

```
$ cat test.sh
#!/bin/sh

export HEY=hey
echo $HEY
```

```
$ ./test.sh
hey
$ echo $HEY

```

output/JupyterLab-Tensorflow/mc-tenant-wrapper.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+# This script checks if the tenant if newer than an exiting one.
+# Every time that the user calls mc, the for loop checks to see if any vault secrets are newer than the most recent lockfile. 
+# If the secret is newer, then it reinitializes the client right away.
+
+for  f in $(ls /vault/secrets/minio-* | grep -v json); do
+ tenant=$(basename "$f" | sed 's/.json$//' | sed 's/^minio-//') # remove the json suffix and minio- prefix 
+ if [ ! -f /tmp/.minio-$tenant ] || [ $f -nt /tmp/.minio-$tenant ]; then
+     source $f
```
brendangadd, line 9: Isn't this going to force the user's env to contain the lexicographically last tenant's variables? If so, as a result, wouldn't the following happen?
* User interacts with minimal tenant
* Creds are updated
* Environment transparently references premium tenant
* User sees no break in connectivity, but doesn't recognize MinIO state

resources/mc-tenant-wrapper.sh
```diff
@@ -0,0 +1,14 @@
...
+
+for  f in $(ls /vault/secrets/minio-* | grep -v json); do
+ tenant=$(basename "$f" | sed 's/.json$//' | sed 's/^minio-//') # remove the json suffix and minio- prefix 
+ if [ ! -f /tmp/.minio-$tenant ] || [ $f -nt /tmp/.minio-$tenant ]; then
+     source $f
+     mc config host add $tenant $MINIO_URL $MINIO_ACCESS_KEY $MINIO_SECRET_KEY
+     touch /tmp/.minio-$tenant
+ fi
+done
+/usr/local/bin/mc-wrapped $@
```
ca-scribner, line 14: this name is outdated.  Sync it up with what is in the `Dockerfile` :)

Nancy-Badran: oops yeah

output/RStudio/mc-tenant-wrapper.sh
Nancy-Badran: @ca-scribner , In fact I am not 100% sur that it will not loop since `config host add ...`  must call the original mc copied to  /usr/local/bin/mc-original. For testing, if I will test it locally in a running container , as you know, I cant access to /vault/secrets/minio-* since the minio credentials are managed by vault and injected via a k8s pod, if I am not wrong, and if I test it in a running notebook server the /usr/local/bin/mc-original will not be there, and must to call it somewhere, so it will not a good way to ensure that the bash script will not loop . Do you have another way to do the test?

ca-scribner: Some ideas for testing:
* build/run the docker image locally
* create your own /vault/secrets/* files (vault would normally mount these files, but they're just text files so you can create them yourself.  Or, you can add a `COPY` at the end of your Dockerfile to debug (then remove it later)
* see what it does

You also don't actually need to access those files.  You can mock that part out of your script (just comment out the things that needed that file at the top of your script during testing).  You need to decide if the `mc` call will recurse or actually get to the original `mc` tool - the `mc` host doesn't need to be real (if you get to the real `mc` tool it'll break saying a host doesn't exist.  If you recurse, you'll never get there)

In general, this is a good case for mocking (eg: faking, like putting your own files in `/vault/*` rather than using real ones mounted by Vault) or commenting out unimportant pieces of the system that you can't easily test with.

Nancy-Badran: Create file with credential and copy it from Dockerfile to /vault/secrets/minio-* . I can tell you it doesn't loop from my side.

ca-scribner (approved):

docker-bits/4_CLI.Dockerfile
```diff
@@ -43,14 +43,11 @@ RUN curl -LO "${KUBECTL_URL}" \
...
   && \
     curl -sLO https://aka.ms/InstallAzureCLIDeb \
     && bash InstallAzureCLIDeb \
     && rm InstallAzureCLIDeb \
   && \
     wget -q "${OH_MY_ZSH_URL}" -O /tmp/oh-my-zsh-install.sh \
     && echo "${OH_MY_ZSH_SHA} /tmp/oh-my-zsh-install.sh" | sha256sum -c
-
-# Have to use the zsh-installer in the entrypoint!
-COPY helpers.sh /tmp/helpers.sh
```
ca-scribner, line 56: I don't think the intent was to remove helpers entirely.  My guess is this comes from a comment from Blair about removing just the mc-related part of the helpers script?

PR #195, opened 2021-01-23
feat(rstudio): upgrade rstudio to 1.4

blairdrummond:
This was more complicated than expected! It also adds a [fork of rsession-proxy](https://github.com/blairdrummond/jupyter-rsession-proxy)

# SQLite permission change (chown)

First error encountered was a sqlite error not being able to write to a database. The issue seems to have been a simple ownership issue. `/var/lib/rstudio-server/rstudio.sqlite` is owned by `rstudio-server`, but when we start rstudio via jupyter, the running user is `jovyan`. So needed to add:

```
chown $NB_USER:users /var/lib/rstudio-server/rstudio.sqlite
```

And *also* need rserver to use the jovyan user, which leads to [this proxy change](https://github.com/blairdrummond/jupyter-rsession-proxy/blob/eb5c4b7023f82ab7841d55fef54c56ebe7a51705/jupyter_rsession_proxy/__init__.py#L38).

# auth-none=1 and www-root-path

This is just a new setting introduced by [this PR](https://github.com/rstudio/rstudio/pull/7628#issuecomment-723190694). Basically ALL rstudio auth mechanisms send you to an auth page, but if auth is none then you proceed immediately... However, since we run on `/rstudio` not `/`, we have to configure the www-root-path or else auth redirects to a non-existent path.

So that's where we get [the proxy changes](https://github.com/blairdrummond/jupyter-rsession-proxy/blob/9faa6d5114b280bf80846c3841a58d78e17d1494/jupyter_rsession_proxy/__init__.py#L34-L45)

**There is discussion of this on [this rsession-proxy issue](https://github.com/jupyterhub/jupyter-rsession-proxy/issues/95#issuecomment-766412024)**

## Viola
Screenshot from 2021-01-24 13-52-50

### DIFF
--- a/docker-bits/6_RStudio.Dockerfile
+++ b/docker-bits/6_RStudio.Dockerfile
@@ -1,10 +1,8 @@
 # install rstudio-server
-ARG RSTUDIO_VERSION=1.1.463
-ARG SHA256=62aafd46f79705ca5db9c629ce3b60bf708d81c06a6b86cc4b417fbaf30691c1
+ARG RSTUDIO_VERSION=1.4.1103
+ARG SHA256=552baf1bbfd98fc36a3f63c430cce3ceb16dd41de723a535fa27d254ff6afa62
 RUN apt-get update && \
-    wget http://archive.ubuntu.com/ubuntu/pool/main/o/openssl1.0/libssl1.0.0_1.0.2n-1ubuntu5.5_amd64.deb -O libssl1.0.0.deb && \
-    dpkg -i libssl1.0.0.deb && \
-    curl --silent -L --fail "https://download2.rstudio.org/rstudio-server-${RSTUDIO_VERSION}-amd64.deb" > /tmp/rstudio.deb && \
+    curl --silent -L  --fail "https://download2.rstudio.org/server/bionic/amd64/rstudio-server-${RSTUDIO_VERSION}-amd64.deb" > /tmp/rstudio.deb && \
     echo "${SHA256} /tmp/rstudio.deb" | sha256sum -c - && \
     apt-get install --no-install-recommends -y /tmp/rstudio.deb && \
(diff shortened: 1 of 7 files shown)


blairdrummond:
Closes https://github.com/StatCan/daaas/issues/332  and  https://github.com/StatCan/daaas/issues/302

blairdrummond:
Linking for the interested: https://github.com/jupyterhub/jupyter-rsession-proxy/issues/95

blairdrummond:
Converted to draft. This PR works on Localhost, but we're still stuck on Kubeflow. Thanks @ryanlovett for catching my mistake

Screenshot from 2021-01-24 08-12-29

blairdrummond:
Fixed now, but would benefit from a review of someone who knows this stuff better. Maybe @brendangadd  if you get the chance?

wg102 (approved):

blairdrummond:
@brendangadd I'd appreciate your take before merging. Tl;dr is 

- [This is the upstream issue](https://github.com/jupyterhub/jupyter-rsession-proxy/issues/95)
- [This is my patch](https://github.com/blairdrummond/jupyter-rsession-proxy/blob/a65a984cad5b9b250eeafd20803d9eb60f81429d/jupyter_rsession_proxy/__init__.py#L32-L44)
-  Then we export `RSTUDIO_WWW_ROOT_PATH` in the docker entrypoint
- Also `chown` a sqlite database in the dockerfile, because it's owned by `rstudio-server` but we run as `jovyan`. (there's a corresponding --server-user in the proxy, too). RStudio was crashing with an authentication thing without this

brendangadd:
@blairdrummond This seems totally reasonable. We can worry about the proxy headers etc. once upstream session-proxy settles on an approach. LGTM!

blairdrummond:
Yeehaw!!!

PR #196, opened 2021-01-26
fix(rstudio): Rstudio hardening

wg102:
fixes https://github.com/StatCan/daaas/issues/352

Need to verify it works as expected. 
Might be needed to :

- add `www-allow-origin`
- Remove `www-frame-origin=none` since it is set in the proxy

### DIFF
--- a/docker-bits/6_RStudio.Dockerfile
+++ b/docker-bits/6_RStudio.Dockerfile
@@ -6,7 +6,12 @@ RUN apt-get update && \
     echo "${SHA256} /tmp/rstudio.deb" | sha256sum -c - && \
     apt-get install --no-install-recommends -y /tmp/rstudio.deb && \
     rm /tmp/rstudio.deb && \
-    apt-get clean && rm -rf /var/lib/apt/lists/*
+    apt-get clean && rm -rf /var/lib/apt/lists/* && \
+    #Harden rstudio-server
+    echo "www-frame-origin=none" >> /etc/rstudio/rserver.conf && \ 
+    echo "www-enable-origin-check=1" >> /etc/rstudio/rserver.conf && \ 
+    echo "www-same-site=lax" >> /etc/rstudio/rserver.conf && \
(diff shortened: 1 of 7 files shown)


blairdrummond:
Image runs great! I can launch shiny dashboards and all that stuff. The image *does* prevent uploads over 100Mb, as one would expect, but the UX of it **sucks**.

Screenshot from 2021-01-28 20-53-34

No warning to the user or anything.

Gonna tag @brendangadd to see what his feelings are, cause I guess it weighs the UX against the AVA scan recommendations. (Unless we patch the message to actually explain to the user why the upload failed).

docker-bits/6_RStudio.Dockerfile
```diff
@@ -6,7 +6,13 @@ RUN apt-get update && \
...
     apt-get install --no-install-recommends -y /tmp/rstudio.deb && \
     rm /tmp/rstudio.deb && \
-    apt-get clean && rm -rf /var/lib/apt/lists/*
+    apt-get clean && rm -rf /var/lib/apt/lists/* && \
+    #Harden rstudio-server
+    echo "www-frame-origin=none" >> /etc/rstudio/rserver.conf && \ 
+    echo "www-enable-origin-check=1" >> /etc/rstudio/rserver.conf && \ 
+    echo "www-same-site=lax" >> /etc/rstudio/rserver.conf && \
+    echo "restrict-directory-view=1" >> /etc/rstudio/rsession.conf && \
+    echo "limit-file-upload-size-mb=100" >> /etc/rstudio/rsession.conf
```
blairdrummond, line 15: Personally think we should remove this upload limit, as it has bad UX and can be handled at the envoy level.

blairdrummond (approved):
Approving, but adding @brendangadd for his take on the upload thing.

blairdrummond:
Resolved in standup.

@wg102  can you remove the upload limit then we'll merge?

PR #206, opened 2021-02-01
update: docker-stack base image version

### LINKED ISSUE #197: Update inherited docker-stacks image version
The version of the image we currently inherit from jupyter/docker-stacks (tag:42f4c82a07ff) in base-notebook is outdated.
New image with Jupyterlab 3.0.5 is now available.

JessicaBarh:
Resolves #197

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -57,6 +57,9 @@ jobs:
         COMMIT=$(make get-commit)
         echo
         cd output/${{ matrix.notebook }}
+        if [[ ${{ matrix.notebook }} == *"OL-compliant" ]]; then
+          COMMIT="r-4.0.3"
+        fi
         docker build . --build-arg BASE_VERSION=$COMMIT -t localhost:5000/kubeflow-image:$SHORT_SHA
         docker push localhost:5000/kubeflow-image:$SHORT_SHA
         docker rmi localhost:5000/kubeflow-image:$SHORT_SHA
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -57,9 +57,6 @@ jobs:
         COMMIT=$(make get-commit)
         echo
         cd output/${{ matrix.notebook }}
-        if [[ ${{ matrix.notebook }} == *"OL-compliant" ]]; then
-          COMMIT="r-4.0.3"
-        fi
         docker build . --build-arg BASE_VERSION=$COMMIT -t localhost:5000/kubeflow-image:$SHORT_SHA
         docker push localhost:5000/kubeflow-image:$SHORT_SHA
         docker rmi localhost:5000/kubeflow-image:$SHORT_SHA
--- a/docker-bits/0_CPU-OL-compliant.Dockerfile
+++ b/docker-bits/0_CPU-OL-compliant.Dockerfile
@@ -1,4 +1,7 @@
 ARG BASE_VERSION=r-4.0.3
+
+# temporarily hard coding docker image tag to r-4.0.3 to avoid using the current tag passed in --build-arg in CI for OL. 
+# Should be replaced back to $BASE_VERSION when current images are replaced for the OL ones 
 FROM jupyter/datascience-notebook:r-4.0.3
 
 USER root
--- a/docker-bits/0_CPU-OL-compliant.Dockerfile
+++ b/docker-bits/0_CPU-OL-compliant.Dockerfile
@@ -1,0 +1,9 @@
+ARG BASE_VERSION=r-4.0.3
+FROM jupyter/datascience-notebook:$BASE_VERSION
+
+USER root
+ENV PATH="/home/jovyan/.local/bin/:${PATH}"
+
+RUN apt-get update --yes \
+    && apt-get install --yes language-pack-fr \
+    && rm -rf /var/lib/apt/lists/*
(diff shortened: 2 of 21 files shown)


JessicaBarh:
using the new Docker-stacks image that includes JupyterLab 3.0, there are now issues in the tensorflow images when installing tensorflow-gpu. It tries to install using channel conda-forge but it should be using "defaults" (pkgs/main/linux-64) see [CI build.](https://github.com/StatCan/kubeflow-containers/runs/1809894941#step:6:825)

running `cat .condarc ` does not show defaults in the list of channels but running `conda config --append channels defaults` returns "'defaults' already in 'channels' list, moving to the bottom"

JessicaBarh:
_tensorflow-gpu_ version installed in our original JupyterLab-Tensorflow image (JLab 2) using conda is outdated - **1.14.0** 
Changed it to use pip to install latest version - **2.4.1** in both OL and non-OL Tensorflow images

.github/workflows/build.yaml
```diff
@@ -57,6 +57,9 @@ jobs:
         COMMIT=$(make get-commit)
         echo
         cd output/${{ matrix.notebook }}
+        if [[ ${{ matrix.notebook }} == *"OL-compliant" ]]; then
+          COMMIT="r-4.0.3"
+        fi
```
ca-scribner, line 62: This would be a temporary change while we maintain two streams, one using an older docker-stacks and one using the newest one?

ca-scribner: If yes, let's add a comment in here (and push.yaml) to reflect that just so it is clear.  Something that states this is temporary and the condition on which we should remove it (that way when we **do** forget to remove it and then notice this in months, we know what to do).  Maybe like "Temporary fix to support ..." or such.  

Let's also combine this with the `COMMIT=` on 57 and add an echo statement in this if block that say we're overriding `COMMIT`.  I'm thinking something like
```
if is OL-compliant; then
  # Temporary special case to address ... (this is the statement where you can say why we did this and when it should be removed, etc)
   ...
else
  # Typical case
  COMMIT=$(make get-commit)
fi
```

docker-bits/0_CPU-OL-compliant.Dockerfile
```diff
@@ -0,0 +1,9 @@
+ARG BASE_VERSION=r-4.0.3
```
ca-scribner, line 1: As an alternative to hard-coding the docker version override in the Makefile, you could also put it here.  You could ignore `BASE_VERSION` and just hard-code to r-4.0.3.  Not sure it is a better way just mentioning it.

docker-bits/2_Tensorflow.Dockerfile
```diff
@@ -1,10 +1,5 @@
 # Install Tensorflow
-RUN conda create -n tensorflow && \
-    conda install -n tensorflow --quiet --yes \
-      'tensorflow-gpu' \
-      'keras' \
-      'ipykernel==5.3.4' \
-    && \
-    conda clean --all -f -y && \
-    fix-permissions $CONDA_DIR && \
-    fix-permissions /home/$NB_USER
```
ca-scribner, line 10: These commands always need to be included no matter what when we `pip` or `conda` install anything.  They change the permissions of the contents of the `conda_dir` (which includes where `pip` installs things) to be writable by the `jovyan` user.  This is important to make python usable (otherwise people can't modify their packages).  

They also have to be in the same `RUN` command (they can't be something like `RUN pip install ... ; RUN fix-permissions`).  If they're not done in the same `RUN`, but instead done later, because of how Docker handles layers it'll actually duplicate all the data!  Because of that, by leaving it out here entirely, the next time someone does a `fix-permissions` somewhere later in the Dockerfiles it'll add a surprise 2-3GB to their image.

docker-bits/3_Kubeflow.Dockerfile
```diff
@@ -1,10 +1,9 @@
 RUN pip --no-cache-dir install --quiet \
       'git+https://github.com/statcan/kubeflow-pipelines@b47c8de7f2915722c5c91bf3b1c7d54b946ef2a6#subdirectory=sdk/python/' \
-      'kfp-server-api==1.2.0' \
-      'kfp-tekton==0.5.0' \
+      'kfp-server-api==1.3.0' \      
       'kubeflow-fairing==1.0.2' \
-      'ml-metadata==0.26.0' \
-      'kubeflow-metadata==0.3.1' \
+      'ml-metadata==0.24.0'\
+      'kubeflow-metadata==0.2.0' \
```
ca-scribner, line 6: Do you know what the error was here that forced this change?  I think reverting breaks something (can't remember what, but I remember **something** breaking ;)  ).  Maybe we can fix the problem and keep the higher versions

ca-scribner (changes requested):

docker-bits/3_Kubeflow.Dockerfile
JessicaBarh: Using kubeflow-metadata==0.3.1 and ml-metadata==0.26.0 gave:
```
ERROR: Cannot install kubeflow-metadata==0.3.1 and ml-metadata==0.26.0 because these package versions have conflicting dependencies.
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
```

JessicaBarh: so ml-metadata==0.27.0 and ml-metadata==0.2.0 works too

docker-bits/0_CPU-OL-compliant.Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+FROM jupyter/datascience-notebook:r-4.0.3
```
ca-scribner, line 2: Please add a comment here noting how this is temporary and overrides the usual behaviour etc etc

ca-scribner (approved):

PR #207, opened 2021-02-01
fix: Extend timeout for Trivy in build.yaml

ca-scribner:
pytorch scan timed out (took longer than 2 min).  Extending timeout to fix


blairdrummond (approved):
LGTM!

PR #220, opened 2021-02-17
feat(jupyterlab): bump main images to 3.0

### LINKED ISSUE #219: upgrade jupyterlab 2 images to jupyterlab 3
use Jupyterlab 3.0 for production images in kubeflow-containers and update manifest to use these.

Note: This is different from the OL-compliant images, some extensions such as jupyterlab-git and jupyterlab-lsp (#218)  that are not OL-compliant will be included.

Eventually we will only have the OL-compliant images once i18n is available for all our desired extensions

JessicaBarh:
Resolves #219

### DIFF
--- a/docker-bits/6_JupyterLab.Dockerfile
+++ b/docker-bits/6_JupyterLab.Dockerfile
@@ -1,4 +1,5 @@
 # installs vscode server, python & conda packages and jupyter lab extensions.
+# TODO: Change jupyterlab-git pre-release install to official v0.30.0 release once available
 
 # Install vscode
 ARG VSCODE_VERSION=3.8.0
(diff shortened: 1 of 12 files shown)


docker-bits/6_JupyterLab.Dockerfile
```diff
@@ -79,6 +70,10 @@ RUN pip install --quiet \
   fix-permissions $CONDA_DIR && \
   fix-permissions /home/$NB_USER
 
+# Install R and Julia language servers
+RUN julia -e 'using Pkg; Pkg.add("LanguageServer")' && \
+    conda install -c conda-forge r-languageserver
```
blairdrummond, line 75: Do we need to clean the layer and/or fix permissions after this?

ca-scribner: Agreed, this will need a fix permissions and a conda clean or whatever the command is to clear the cache

JessicaBarh: Sounds good, I'll add that in.

ca-scribner (approved):
lgtm!

PR #232, opened 2021-03-10
fix: Handle JL2 config breaking JL3, update stale file paths

ca-scribner:
closes statcan/daaas#464 and #233 

This PR:
1. Fixes a bug where an incompatible setting from JL2 causes a JL3 notebook to go into a crash loop
2. Fixes the pathing on several `./resource` files.  These files were moved in previous commits but never updated in their `Dockerfile`s.  This was not caught because `make clean` was not run, so the old files in the original locations were never deleted.  When doing this PR, running `make clean` at the start made these breaks apparent

Most changed files are to address (2.): a few files were moved, which causes a lot of apparent changes in `./output`

To address (1.), `resources/common/start_custom.sh` was edited

To test the fix for (1.):
* Create a new notebook server and edit/create the `.jupyter/jupyter_notebook_config.json` (see Statcan/daaas#464) to include (this creates a config file that jl2 is fine with but jl3 will break on):
```
{
    "NotebookApp": {
        "server_extensions": [
            "jupyter_tooling.tooling_handler"
        ]
    }
}
```
* Shut down the notebook server
* Create a new notebook server using a JL3 image (current jupyterlab-cpu image is a good example).  For workspace volume, reuse the above workspace with the JL3 incompatible config file.  
  * This notebook should go into a crash loop and never fully start (you can see it from `kubectl get pods`.  After a few minutes the `restarts` will tick up).  
* Create a new notebook server using the custom image `jupyterlab-cpu:fix-jl2-to-3-migration-bug4` (that image is the same as in this PR, but was pushed manually to save some time).  For workspace volume, reuse the above workspace with the JL3 incompatible config file
  * This notebook should open up like normal without error


ca-scribner:
Tested by:
* Create a new notebook server and edit/create the `.jupyter/jupyter_notebook_config.json` (see Statcan/daaas#464) to include (this creates a config file that jl2 is fine with but jl3 will break on):
```
{
    "NotebookApp": {
        "server_extensions": [
            "jupyter_tooling.tooling_handler"
        ]
    }
}
```
* Shut down the notebook server
* Create a new notebook server using the custom image `jupyterlab-cpu:fix-jl2-to-3-migration-bug4` (that image is the same as in this PR, but was pushed manually to save some time).  For workspace volume, reuse the above workspace with the JL3 incompatible config file
* JL3 server should open up like normal without error

JessicaBarh (approved):
LGTM!
followed the steps in the description and all works well.

PR #234, opened 2021-03-12
fix:broken vscode python extension

Ito-Matsuda:
Use version v2020.5.86806
Add override to prevent code-server from
 automatically updating the python extension
Pin vs-code extension versions (obtained from current image).

Closes https://github.com/StatCan/daaas/issues/463


ca-scribner (approved):
Build failed but just randomly in the RStudio image, but think this was just an internet problem(?) so lgtm

PR #266, opened 2021-06-01
fix: use xdg-settings to make firefox default

### LINKED ISSUE #265: Default browser is invalid
Launching the default web browser from the dock at the bottom of the desktop produces an error. The default browser is not set to Firefox.

image

blairdrummond:
Good catch. I never use the doc anymore so didn't notice :( 

Closes #265


Ito-Matsuda (approved):
LGTM. Did a quick check to confirm

PR #349, opened 2022-05-10
DO NOT MERGE: testing persistent conda

Jose-Matsuda:
Things we want to do:
--> Confirm what action we want to take

Related to: https://github.com/StatCan/daaas/issues/285

Moved some configuration around, need to test on dev.
Commented out other images so they dont get pushed

Maybe should also do 
https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#specifying-a-location-for-an-environment

PR #356, opened 2022-06-08
Don't enable pip user installs globally

ToucheSir:
# Description

This has been messing with tools that use pip in a virtualenv, e.g. https://github.com/pre-commit/pre-commit/issues/1151. I'm inclined to agree that it promotes bad practices too.

cc @asolis

# Tests / Quality Checks

## Are there breaking changes?
Ask yourself the next question;
- [ ] Do we want to maintain the previous image from which we had to do breaking changes from?

If no, then carry on. If yes, there is a breaking change and we **want to maintain the previous image** do the following
- [ ] Create a new branch for the current version (ex v1) based off the current master/main branch
- [ ]  Increment the tag in the CI for pushes to master/main (v1 to v2)
- [ ] Change the CI that on pushes to the newly created "v1" branch (the name of the newly created branch we want to maintain is) it will push to the ACR. 
## Automated Testing/build and deployment
- [ ] Does the image pass CI successfully (build, pass vulnerability scan, and pass automated test suite)?
- [ ] If new features are added (new image, new binary, etc), have new automated tests been added to cover these?
- [ ] If new features are added that require in-cluster testing (e.g. a new feature that needs to interact with kubernetes), have you added the `auto-deploy` tag to the PR before pushing in order to build and push the image to ACR so you can test it in cluster as a custom image?

## JupyterLab extensions

- [ ] Are all extensions "enabled" (`jupyter labextension list` from inside the notebook)?

## VS Code tests

- [ ] Does VS Code open?
- [ ] Can you install extensions?

## Code review

- [ ] Have you added the `auto-deploy` tag to your PR before your most recent push to this repo?  This causes CI to build the image and push to our ACR, letting reviewers access the built image without having to create it themselves

### DIFF
--- a/resources/common/pip.conf
+++ b/resources/common/pip.conf
@@ -1,6 +1,3 @@
-[install]
-user = true
-
 [global]
 trusted-host = jfrog-platform-artifactory-ha.jfrog-system
 index-url = http://jfrog-platform-artifactory-ha.jfrog-system:8081/artifactory/api/pypi/pypi-remote/simple


Jose-Matsuda:
@ToucheSir you need to run `make generate-dockerfiles` at this repository's root thats why the pre-build-check failed

ToucheSir:
Oh, I don't have a local dev setup for this and just created a diff using the web UI. My hope was to get some feedback/possible objections for this change first, since it seems the config option was very intentionally added in the first place.

vexingly:
Some notes for the discussion, based on my understanding so far...

Most likely the configuration which forces pip installs to a user directory was added because users don't have root/sudo access and the default otherwise is a system folder - which they would be denied to install their package within.

It does make sense as a best practice to install python packages in a virtual environment, which is why Conda is recommended (as well as the py_install wrapper). However it seems the user flag does not work well with py_install at least, and needs to be manually disabled. Making it disabled by default could increase the number of support requests with users trying to use pip, not sure if Conda is impacted either way.

ToucheSir:
The issue is that flipping this flag on globally blocks more than it enables. If there were an equivalent `--no-user`, then it would be less of a problem. As-is though, it is impossible to:
- Install packages in a virtualenv
- Use tools which create venvs for their dependencies (which is a problem for adoption of our [data science project template](https://github.com/StatCan/datascience-cookiecutter/))

So on balance, it seems easier to direct users to the RPUG documentation (which suggests `--user` as a best practice for global installs) than to tell anyone who needs a virtualenv to create a random config file on every notebook server.

vexingly:
Just trying to make sure we cover all of our bases @ToucheSir, I did find another comment that might be relevant to how things are currently configured:

```
With autoscaling enabled, automated shutdown/restart of notebooks might have some unexpected consequences for users because of how the notebook image is stateless and conda/pip installs will by default go there. Everything would be as expected if you use packages from your notebook server's image, but installing your own causes issue. For example, this would result in an error:

create notebook server
pip install somepackage
conda install somepackage2
pause notebook
resume notebook
use packages somepackage and somepackage2
We've mitigated part of this by forcing a --user on all pip install commands (via pip.conf), which means pip packages go on the stateful workspace volume, but haven't come up with a good solution for conda. The best workaround so far is to only use packages that are in your notebook image.
```

This seems to indicate if you disable user installs and install your custom packages, if your notebook is 'stopped' from the kubeflow UI, when it resumes the packages may no longer be installed. I'm not sure if this is still the case, but something we should validate before making changes.

ToucheSir:
Interesting finding. Is it even possible to run `pip install` on the notebook without `--user` (or the corresponding config option) though? I just assumed it would fail due to lack of permissions. The part about conda sounds like what https://github.com/StatCan/aaw-kubeflow-containers/pull/349 is trying to address?

ToucheSir:
Looks like this was fixed by https://github.com/StatCan/aaw-kubeflow-containers/pull/422 :)

PR #357, opened 2022-06-10
chore(sas): re-enable and update the base

Jose-Matsuda:
Thanks to efforts of Will can re-enable this.
Closes https://github.com/StatCan/daaas-private/issues/11

### DIFF
--- a/.github/workflows/build_push.yaml
+++ b/.github/workflows/build_push.yaml
@@ -66,7 +66,7 @@ jobs:
           # TODO: Pull this from a settings file or Makefile, that way Make can have the same list
           #- docker-stacks-datascience-notebook # Debugging
           - rstudio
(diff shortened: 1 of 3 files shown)


rohank07 (approved):

Jose-Matsuda:
It seems like there's an intermediate image or something that is making the SAS build fail.

Here's what my `docker images` looks like
image

as you can see there is an untagged 13.1 GB image just sitting there, after I did a `docker system prune` to remove it, and then a `docker build` it went back to the beginning of the `Dockerfile` not using the present `test` image that is already there. So yeah I guess at the `6_sas.Dockerfile` step it says sayonarra to the old one and works off of it

Jose-Matsuda:
I will try merging and stopping one of the other workflows just to see what happens

Issue #380, opened 2022-08-17
Azure Container Scan fails due to unlisted Vulnerabilities


Souheil-Yazji:
After merging in https://github.com/StatCan/aaw-kubeflow-containers/pull/379 to resolve https://github.com/StatCan/daaas-private/issues/16 the remote image build fails with the following 

```
/usr/bin/tar xz --warning=no-unknown-keyword -C /home/runner/work/_temp/3eda97c1-b67c-4030-8c67-1244e3d81138 -f /home/runner/work/aaw-kubeflow-containers/aaw-kubeflow-containers/_temp/tools/trivy
Scanning for vulnerabilties in image: localhost:5000/remote-desktop:null
╔══════════════════════╤═════════════════╤═════════════════╤════════════════════════════════════════════════════╤══════════════════════╗
║ VULNERABILITY ID     │ PACKAGE NAME    │ SEVERITY        │ DESCRIPTION                                        │ TARGET               ║
╚══════════════════════╧═════════════════╧═════════════════╧════════════════════════════════════════════════════╧══════════════════════╝

Creating scan result. image_name: localhost:5000/remote-desktop:null, head_sha: 595c080a0f142df4ca1a17ff10188fec8933e4ab
Looks like the scanitizer app is not installed on the repo. Falling back to check run creation through GitHub actions app...
For a better experience with managing allowedlist, install Scanitizer app from https://github.com/apps/scanitizer.
Creating check run. Name: [container-scan] localhost:5000/remote-desktop:null, head_sha: 595c080a0f142df4ca1a17ff10188fec8933e4ab
Created check run. Url: https://github.com/StatCan/aaw-kubeflow-containers/runs/7811713796
Error: Vulnerabilities were detected in the container image
```
see [Githab action](https://github.com/StatCan/aaw-kubeflow-containers/runs/7810977764?check_suite_focus=true)

No vulnerabilities are shown, possible cause is vuln detected in trivy but unexpected behaviour is occurring when container-scan is handling the whitelisted values. 
Update requested in upstream repo https://github.com/Azure/container-scan/issues/143

Temporary fix at #381

Souheil-Yazji:
Update: 
Checking whether trivy whitelisting is functional:
Starting with a trivy scan:
`trivy image -f json -o results.json k8scc01covidacr.azurecr.io/remote-desktop:qgis-key-update --security-checks vuln --timeout=20m`

then created the following .trivyignore file from the allowedlist.yaml file:
[...]

PR #382, opened 2022-08-17
feat(trivy): replace Azure container scan with trivy

Souheil-Yazji:
# Description

Replaces Azure Container Scan with trivy. Note that dockle wasn't being used in the Azure container scan (run-quality-checks: false)

Closes https://github.com/StatCan/aaw-kubeflow-containers/issues/380 
# Tests / Quality Checks

## Code review

- [x] Have you added the `auto-deploy` tag to your PR before your most recent push to this repo?  This causes CI to build the image and push to our ACR, letting reviewers access the built image without having to create it themselves


.github/workflows/build_push.yaml
```diff
@@ -159,14 +159,22 @@ jobs:
...
-    # - uses: Azure/container-scan@v0
+    # - uses: Azure/container-scan@v0.1
     #   if: steps.notebook-name.outputs.NOTEBOOK_NAME != 'sas'
     #   env:
     #     TRIVY_TIMEOUT: 30m0s  # Trivy default is 2min.  Some images take a bit longer
     #   with:
     #     image-name: ${{ steps.build-image.outputs.full_image_name }}
     #     severity-threshold: CRITICAL
     #     run-quality-checks: false
+    #     trivy-version: "0.29.1"
```
brendangadd, line 170: If we're leaving Azure container-scan, then this block won't be coming back; in which case, remove it.

.github/workflows/build_push.yaml
```diff
@@ -159,14 +159,22 @@ jobs:
...
+
+    - name: Aqua Security Trivy
+      if: steps.notebook-name.outputs.NOTEBOOK_NAME != 'sas'
```
brendangadd, line 173: Why not scan the SAS image?

.github/workflows/build_push.yaml
```diff
@@ -159,14 +159,22 @@ jobs:
...
+      run: |
+        curl -sfL https://raw.githubusercontent.com/aquasecurity/trivy/main/contrib/install.sh | sh -s -- -b /usr/local/bin latest
```
brendangadd, line 175: Checksum the binary before using it, and pin a version to make things reproducible (and not fail the checksum).

brendangadd (changes requested):

.github/workflows/build_push.yaml
Souheil-Yazji: 2 critical CVEs are detected in the SAS image, but apparently they were mitigated at some point. I don't agree with the sas image going unscanned but I am limiting the scope of this ticket and leaving the prior decision of not scanning it in. We can address this in another small pr. 
https://github.com/StatCan/daaas-private/issues/11

.github/workflows/build_push.yaml
Souheil-Yazji: The binary is checksum-ed via the install.sh script I'm pulling here. I guess we can checksum the install.sh since that's what we're pulling from remote? see `hash_sha256_verify()` in https://raw.githubusercontent.com/aquasecurity/trivy/main/contrib/install.sh

brendangadd (approved):

PR #390, opened 2022-09-01
Update jfrog url

Jose-Matsuda:
In the move from the beta 0.4.1 jfrog-platform chart to the GA 10.x.y chart they renamed `artifactory-ha` to just `artifactory`, as a result the `name` of the `Service` has been renamed. This should only be merged when the update to the GA chart is completed and working.

We do not want to accidentally have to restart everyone's workloads only to have them unable to access artifactory because the move to the GA chart was not yet completed.


Souheil-Yazji (approved):
LGTM

Issue #418, opened 2022-11-24
Create env variable in startup script to facilitate testing


Jose-Matsuda:
From this issue https://github.com/StatCan/aaw-kubeflow-containers/issues/417

The idea came up that when testing things we often need to push to the acr; but especially when the pod is in a crashloopbackoff and we cannot shell in.

Could have an env variable inside the startup script that if that environment variable exists then we just have control and lets us shell in .

Could have this line for example; https://stackoverflow.com/a/22100106 `sleep infinity` might work 
other options https://stackoverflow.com/questions/45461263/how-can-i-keep-docker-container-running
https://stackoverflow.com/questions/25775266/how-to-keep-docker-container-running-after-starting-services

Issue #417, opened 2022-11-24
User notebook exits as soon as it starts up


Jose-Matsuda:
## User changes affecting `jupyter server` and causing it to break

There is a user in prod whose notebook continues to die;
Made some changes to the `start-custom.sh` to include tracing statements.

![image](https://user-images.githubusercontent.com/23174198/203822714-e53054e2-fb07-419b-8403-c68d3600857c.png)

So it really dies out after `/opt/conda/bin/jupyter server --notebook-dir=/home/${NB_USER} `

Worst case scenario (maybe Monday) can fix this specific instance, and then ask to reproduce in a different server to let me mess around with, as maybe the bug fix is a bit bigger / out of scope of this initial reporting.
I'm making the executive decision to go for the fix this server now in hopes of having it available for use. Will also make a pr that would hopefully enforce virtual environments to avoid this in the future, (its good practice anyways).

### Prevention Strats 
Maybe make use of a python [venv](https://www.geeksforgeeks.org/create-virtual-environment-using-venv-python/). Either way thats not super obvious, still looking for a way to reset things....
Could also [force them to create a venv](https://docs.python-guide.org/dev/pip-virtualenv/)

Jose-Matsuda:
User reported that he might have messed with pip global settings, but when I do `whereis pip` it points to _non-persisted directories_, just like with `conda` / when we do the `/opt/conda... jupyter server`. With my initial thoughts of because these are not persisted (and checking the spec the user did not change his volume to persist anything under /opt/) the simple stop and start should have fixed it but it did not.

Jose-Matsuda:
Maybe user did change well somethign since the last line is `ImportError: cannot import name 'soft_unicode' from 'markupsafe' (/home/jovyan/.local/lib/python3.8/site-packages/markupsafe/__init__.py) `

Jose-Matsuda:
Possibly related https://github.com/pallets/markupsafe/issues/282

Jose-Matsuda:
Been giving this thought, so since the problem appears to be with the mounted volume and not the image itself, would it make sense to put Julian on a different image tag so we can freely mess around in the container? We'd need to change the entrypoint to not run the jupyter server so it doesnt die and we cant shell in to fix it.

That or in the startup script we override and force pip to NOT USE whatever is in /home/jovyan and instead use the /opt/conda installs

Jose-Matsuda:
## DEFAULTS ON A NEW NOTEBOOK

[...]

PR #419, opened 2022-11-25
add sleep on env, force venv, test

Jose-Matsuda:
Related to https://github.com/StatCan/aaw-kubeflow-containers/issues/417 (force users to create and use a venv)
Add a sleep depending on if an environment variable is present https://github.com/StatCan/aaw-kubeflow-containers/issues/418

### DIFF
--- a/.github/workflows/build_push.yaml
+++ b/.github/workflows/build_push.yaml
@@ -66,12 +66,12 @@ jobs:
         notebook:
           # TODO: Pull this from a settings file or Makefile, that way Make can have the same list
           # - docker-stacks-datascience-notebook # Debugging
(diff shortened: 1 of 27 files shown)


Jose-Matsuda:
Sleep seems to work well (notebook did not connect because well it slept, but could exec into it)
image
image

Jose-Matsuda:
I am now checking by reverting the VScode changes. If that is well will uncomment the workflows etc.

Jose-Matsuda:
To create a virtual environment (without pip installing anything)

`python3 -m venv /path/to/new/virtual/environment`
then
`source virtualenv_name/bin/activate`

PR #455, opened 2023-02-28
update(remote-desktop): $XDG_DATA_HOME -> /etc/share

bryanpaget:
# Description

This is a fix for https://github.com/StatCan/daaas/issues/1535, where we needed to change the location of the user's VSCode extensions and settings to avoid data loss due to overwriting `/home/jovyan` whenever Kubeflow workspace volumes are mounted.


bryanpaget:
My initial attempt at fixing this bug has not panned out as expected. Perhaps I just didn't do the due diligence. 

The only change I made for this first attempt was adding:

`ENV XDG_DATA_HOME=/etc/share`

to the Ubuntu Desktop Dockerfile.

bryanpaget:
I have achieved something resembling success, setting the undocumented environmental variable `VSCODE_EXTENSIONS` has enabled the extensions to show up on first boot.

image

Found here in the source code:

https://github.com/microsoft/vscode/blob/64f8dd4634af8349176bc9b1a2080c5cbf3e9e17/src/vs/platform/environment/common/environmentService.ts#L142

docker-bits/0_cpu.Dockerfile
```diff
@@ -13,3 +13,9 @@ ENV PATH="/home/jovyan/.local/bin/:${PATH}"
 RUN apt-get update --yes \
     && apt-get install --yes language-pack-fr \
     && rm -rf /var/lib/apt/lists/*
+
+#updates package to fix CVE-2023-0286 https://github.com/StatCan/daaas-private/issues/57
+#TODO: Evaluate if this is still necessary when updating the base image
+RUN pip install --force-reinstall cryptography==39.0.1 && \
+   fix-permissions $CONDA_DIR && \
+   fix-permissions /home/$NB_USER
```
wg102, line 21: Missing end of file newline

docker-bits/0_cpu_sas.Dockerfile
```diff
@@ -15,3 +15,9 @@ ENV PATH="/home/jovyan/.local/bin/:${PATH}"
 RUN apt-get update --yes \
     && apt-get install --yes language-pack-fr \
     && rm -rf /var/lib/apt/lists/*
+
+#updates package to fix CVE-2023-0286 https://github.com/StatCan/daaas-private/issues/57
+#TODO: Evaluate if this is still necessary when updating the base image
+RUN pip install --force-reinstall cryptography==39.0.1 && \
+   fix-permissions $CONDA_DIR && \
+   fix-permissions /home/$NB_USER
```
wg102, line 23: Missing end of file newline

trigger.md
```diff
@@ -0,0 +1 @@
+hello
```
wg102, line 1: ...Why?

wg102 (changes requested):

trigger.md
Souheil-Yazji: I think maybe he was trying to trigger the CI, should be removed.

docker-bits/0_cpu.Dockerfile
Souheil-Yazji: These already exist on master. Is the branch history up to date?

Souheil-Yazji:
image
Initial review of `remote-desktop:5c67a4f6` with default volume mount appears to load up the extensions successfully. It seems however that there's consistent crashing/dead loading of extenstions (they never load and process keeps spinning) from vscode. Investigating further.

bryanpaget:
> Initial review of `remote-desktop:5c67a4f6` with default volume mount appears to load up the extensions successfully. It seems however that there's consistent crashing/dead loading of extenstions (they never load and process keeps spinning) from vscode. Investigating further.
> 
> image

I noticed this as well.

> 
> Additional observations from testing:
> 
> VSCODE takes a little while to load, and seems to have to "reinstall" or "reinitialize" the extensions. This could be due to there being two folders of interest: extensions and data. I'm thinking the data folder is still being wiped meaning some data pertaining to the extensions are not being kept and then it has to reinstall the extensions.
> Reusing volumes (previously created workspace and data volumes) works but the above issue persists where you have to wait a couple of minutes for the extensions to reinitialize.
> When you first boot up VSCODE you can see extension buttons on the sidebar, then they disappear, then they are slowly reinstalled.
> Aside from having to wait a minute for the extensions to load this seems to fix the issue.

https://github.com/StatCan/daaas/issues/1535#issuecomment-1457036213

bryanpaget:
> Getting @Jose-Matsuda to confirm behaviour as well. Feel free to test on your end @bryanpaget

I noticed this as well but it's a problem with the normal remote-desktop image in production as well. Not limited to this image.

Jose-Matsuda:
Yeah I'm getting it as well.
The below happens on an unclassified remote desktop 1,4,16.
image

For the prob remote desktop same settings it hasnt given the above error but I do oddly only see two installed extensions
image

I'll take a look at the behaviour of the extensions without a volume attached on our current v1 remote desktop to see if we can get the 133 error

bryanpaget:
> Yeah I'm getting it as well. The below happens on an unclassified remote desktop 1,4,16. image
> 
> For the prob remote desktop same settings it hasnt given the above error but I do oddly only see two installed extensions image
> 
> I'll take a look at the behaviour of the extensions without a volume attached on our current v1 remote desktop to see if we can get the 133 error

Here's the crash using the image in production, on dev, with a data volume attached. I tried to reproduce without a data volume but it didn't crash. Adding a data volume made it crash this time.

image

bryanpaget:
And here's the 133 on a regular production image on dev without a data volume:

image

Proof:

image

wg102 (approved):

Issue #326, opened 2022-02-23
Slack notification if build fails


blairdrummond:
@chuckbelisle Can we get a slack notification configured when the nightly build fails in this repo? Apparently the nightly build is failing, meaning we have critical CVEs :|

Souheil-Yazji:
Sliding this issue in as this is a pretty common occurance, and we don't find out until someone checks the Github Actions.

menmarc:
@chuckbelisle In which Slack workspace and channel should I send the notification ? 
@Jose-Matsuda Is there a particular content template I should follow regarding the content of the notification ?

chuckbelisle:
It would be more practical to have it on the stats-cloud.slack.com space, but might require the Cloud team to give us permission to access the API/webhooks? If we do already have access to these, we could have the notifications go to a private channel that includes the AAW dev team members.

menmarc:
@chuckbelisle I did a quick test using https://github.com/act10ns/slack for sending Slack notifs to github and it works fine using "incoming-webhook" slack app. 

As for the channel name, should we call it "Kubeflow notifications" ?

menmarc:
@Souheil-Yazji After spoking with Charles he mentioned the channel name should be aaw-kf-build. It is currently setup I'll add you. 
For failure message, it will always occur at the same time and you already have it in Slack so is there a point to adding it to the message? It's easy to add I'm just wondering if it's necessary. Regarding the error message itself, I don't think it's possible to catch unexpected errors at least from what I found. 

The ticket didn't mentioned adding a notif for a success. Again not hard to do but is it required ? Would it create noise in the slack channel ?

Jose-Matsuda:
Personally I dont think we should have notifications for successes, we only care about it if fails

menmarc:
i've merged my change this morning (sorry I though I did it yesterday). I'll keep an eye on the build_push workflow when it fails to see if it right the proper notif in slack.

PR #460, opened 2023-03-02
Add slack notifications for nightly build

menmarc:
Added slack notification for the 2 declared jobs. It will send a slack notification only if the schedule event (night build) fails (aka a failure occurs in declared jobs).

closes https://github.com/StatCan/aaw-kubeflow-containers/issues/326

### DIFF
--- a/.github/workflows/build_push.yaml
+++ b/.github/workflows/build_push.yaml
@@ -58,7 +58,7 @@ jobs:
       uses: act10ns/slack@v1
       with: 
         status: failure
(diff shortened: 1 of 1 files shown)


Jose-Matsuda (changes requested):
Probably don't want the URL to be disclosure-vetting app.

I also think we don't really need it in the pre-build-checks job since there shouldnt be any pushes to the main branch that don't pass this step

Jose-Matsuda (approved):

PR #488, opened 2023-07-20
feat: update code-server and extensions in jupyterlab

vexingly:
Add the databricks and some other useful extensions by default, bump up the code-server version to latest as required by the databricks extension. The latest databricks extension is needed to configure storage accounts as an output location!


vexingly:
Changes in current PR:
- Add cache step to workflow to try to speed up builds (untested)
- Add mamba library to speed up conda env setup
- - Use mamba for conda env setup
- Fix global pip timeout not getting persisted to reduce error messages from artifactory
- Fix AMD AOCL libraries not getting setup
- Update VS Code to latest version in jupyterlab image (remote-desktop is currently being rebuilt)
- Update default extensions that we install (new versions and new extensions)
- Fix process to install extensions and default user settings (move directory back under /home/jovyan/)
- Add R language server install to get rid of VS Code nag message on startup
- Update default theme to new VS Code default (otherwise it nags you on first startup)

PR #491, opened 2023-07-26
Update build_push.yaml: deprecate set-output

bryanpaget:
update workflow to use $GITHUB_OUTPUT instead of set-output.

See: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


Jose-Matsuda (approved):
lgtm

PR #501, opened 2023-08-14
update set-output to GITHUB_OUTPUT

bryanpaget:
Previously I [deprecated set-output for the Github Actions](https://github.com/StatCan/aaw-kubeflow-containers/pull/491), this PR is for deprecating set-output in the Makefile.


mathis-marcotte (approved):
LGTM
output seems to match what we had, but without the deprecation warnings.

new:
image

old:
image

PR #520, opened 2023-09-06
Make torch available

bryanpaget:


bryanpaget:
# Before (on Prod):

image
Source: https://gist.github.com/bryanpaget/c88ec337f6c5d35187f979c6b7f27e59#file-cuda-not-available-ipynb

# After (on Dev):

image
Source: https://gist.github.com/bryanpaget/c88ec337f6c5d35187f979c6b7f27e59#file-cuda-available-ipynb

Makefile
bryanpaget, line 1: I removed the CUDA drivers from the PyTorch and Tensorflow Dockerbits because those are handled by `pytorch-cuda=11.8` in `pytorch`.

Also the rstudio-server docker-bit has been added due to rstudio-server being broken out into its own docker-bit.

docker-bits/2_pytorch.Dockerfile
bryanpaget, line 1: These changes are just to update the torch virtual environment, install the required pacakges including `pytorch-cuda=11.8` which handles the CUDA drivers.

bryanpaget:
Since I removed the manual CUDA installation in favor of the Anaconda packaging, I also had to fix up the Tensorflow image to install cuda in the same way.

bryanpaget:
> If the tensorflow stuff works out with Anaconda, then we can probably do away with the tensorflow image and pytorch image dichotomy and just have separate conda environments on the same jupyterlab-gpu image.

This doesn't seem to work. I can get PyTorch to work with Anaconda's GPU-related packages, but Tensorflow doesn't seem to like the Anaconda GPU-related packages.

Jose-Matsuda (approved):
I see you got caught between a rock and a hard place here, having to update to 2204 in this branch so that when the other branch got merged in this pr wouldnt become obsolete.

Though if I could nitpick id say the remote desktop / rstudio changes are a bit much to have in this PR just because it's not directly related to the purpose of the PR and I guess depending on what gets merged first it might be "lost" anyways since it would be the same as on the main branch.

bryanpaget:
> I see you got caught between a rock and a hard place here, having to update to 2204 in this branch so that when the other branch got merged in this pr wouldnt become obsolete.
> 
> Though if I could nitpick id say the remote desktop / rstudio changes are a bit much to have in this PR just because it's not directly related to the purpose of the PR and I guess depending on what gets merged first it might be "lost" anyways since it would be the same as on the main branch.

Good point. I will modify my PR so that it is more focused.

tests/jupyterlab-tensorflow/test_tensorflow.py
```diff
@@ -23,7 +23,7 @@
 def test_tensorflow(container, name, command):
     """Basic tensorflow tests"""
     LOGGER.info(f"Testing tensorflow: {name} ...")
-    c = container.run(tty=True, command=["start.sh", "python", "-c", command])
+    c = container.run(tty=True, command=["start.sh", "conda", "run", "-n", "tensorflow", "python", "-c", command])
```
bryanpaget, line 26: revert to what is on main

bryanpaget:
### Testing

#### PyTorch

PyTorch seems to work!

image

#### Tensorflow

I'm testing Tensorflow just to make sure I didn't create any issues for that image since PyTorch and Tensorflow were using the same NVIDIA drivers before the update.

image

PR #552, opened 2023-11-15
Update version of pyarrow

wg102:
Closes https://github.com/StatCan/aaw-private/issues/151


Jose-Matsuda:
Seems like that file isn't used / should be removed / it makes sense why it didnt auto-fail at the `pre-check` where you should have hit more than 1 file

Jose-Matsuda (approved):

PR #575, opened 2024-01-23
Ompp cleanup models

Souheil-Yazji:
No longer copying the sample models from the OMPP release to the buckets, this should also speed up the `start-oms.sh` script. If needed, those sample models can still be copied over manually.

### DIFF
--- a/output/docker-stacks-datascience-notebook/start-oms.sh
+++ b/output/docker-stacks-datascience-notebook/start-oms.sh
@@ -66,7 +66,7 @@ if [ ! -d "$OMS_LOG_DIR" ]; then
 fi
 
 # Copy sample models from openmpp installation archive into models directory:
-cp -r "$OMPP_INSTALL_DIR/models/." "$OMS_MODEL_DIR"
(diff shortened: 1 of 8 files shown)


KrisWilliamson (approved):
Looks good.

vexingly:
Any downside for a new namespace that will not have any models available? The web interface would just be empty and users will be responsible to compile or manually copy models before they can run/test? It does seem like for the UAT the users were only interested in utilizing their own models and not the sample models.

jacek-dudek (approved):
I'm ok with them not being there. But maybe it would be good to have one sample model in there by default so that when the openm ui starts it's not completely empty.

Souheil-Yazji:
> Any downside for a new namespace that will not have any models available? The web interface would just be empty and users will be responsible to compile or manually copy models before they can run/test? It does seem like for the UAT the users were only interested in utilizing their own models and not the sample models.

no downside as the folders will still be created, which means the Dir structure is preserved. This will improve the UX for users.

> I'm ok with them not being there. But maybe it would be good to have one sample model in there by default so that when the openm ui starts it's not completely empty.

Honestly, from the perspective of the user, once they have their model in the bucket, it will never be empty.

PR #582, opened 2024-01-30
Fix curl in RStudio caused by openssl version mismatch.

### LINKED ISSUE #583: Investigate curl and rstudio
It appears we have an issue with the installed versions of openssl. According to my research, RStudio wants to use the OS version of openssl but the R packages we have installed with conda want to use the openssl package installed by conda.

```
  call: dyn.load(file, DLLpath = DLLpath, ...)
  error: unable to load shared object '/home/jovyan/R/r-packages-4.3.2/curl/libs/curl.so':
  /lib/x86_64-linux-gnu/libssl.so.3: version `OPENSSL_3.2.0' not found (required by /opt/conda/lib/libcurl.so.4)
```

### References:

Someone else has reported the same issue about 1 year ago as of this writing and they failed to resolve the issue.

- https://discourse.jupyter.org/t/openssl-mismatch-between-rstudio-and-conda-environments/14123

bryanpaget:
/closes #583

### DIFF
--- a/docker-bits/6_rstudio-server.Dockerfile
+++ b/docker-bits/6_rstudio-server.Dockerfile
@@ -8,9 +8,6 @@ RUN apt-get update && \
     apt install -y --no-install-recommends r-base r-base-core r-recommended r-base-dev && \
     apt-get update && apt-get -y dist-upgrade
 
-# https://github.com/rstudio/rstudio/issues/14060
-RUN echo "rsession-ld-library-path=/opt/conda/lib" >> /etc/rstudio/rserver.conf 
-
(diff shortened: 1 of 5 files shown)


docker-bits/6_rstudio-server.Dockerfile
bryanpaget, line 1: I'll leave the updated RStudio version. We didn't need to do this but I personally like keeping up to date.

I also removed the libssl1.1 custom deb install since it doesn't seem to be necessary anymore.

vexingly (approved):
LGTM!

PR #600, opened 2024-03-19
Update jfrog for Cloud Artifactory

wg102:
Related to https://jirab.statcan.ca/browse/BTIS-200

### DIFF
--- a/resources/common/start-custom.sh
+++ b/resources/common/start-custom.sh
@@ -139,16 +139,16 @@ fi
 
 # aaw-dev override settings
 if [[ "$KUBERNETES_SERVICE_HOST" =~ ".131." ]]; then
-  echo "Updating jfrog package config for Dev envrionment"
+  echo "Updating cloud artifactory package config for Dev envrionment"
   
-  pip config --user set global.index-url https://jfrog.aaw.cloud.statcan.ca/artifactory/api/pypi/pypi-remote/simple
+  pip config --user set global.index-url https://artifactory.cloud.statcan.ca/artifactory/pypi-remote/
 
   # remove existing channels in conda system config file
   rm /opt/conda/.condarc
 
-  conda config --add channels https://jfrog.aaw.cloud.statcan.ca/artifactory/api/conda/conda-forge-remote
-  conda config --add channels https://jfrog.aaw.cloud.statcan.ca/artifactory/api/conda/conda-forge-nvidia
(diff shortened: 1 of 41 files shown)


Jose-Matsuda (changes requested):
All the urls are right I believe, but the problem is that cloud artifactory is _not configured_ for anonymous access. 

For example, if you go to the [conda-forge-remote](https://artifactory.cloud.statcan.ca/ui/repos/tree/General/conda-forge-remote) and hit the `Set Me Up` you will see that you have a token generated. 

So the added complexity to this is needing to pass in creds to each config. Similar to how rpug does their setup like [here](https://rpug.pages.cloud.statcan.ca/en/news/2023/6/moving-to-cloud-artifactory.html#proposed-python-configuration) where they have under their `channels`
```
https://svc-ins-rpug:[redacted-token]@artifactory.cloud.statcan.ca/artifactory/api/conda/conda-forge-remote
```

We could pass this in at startup time where it creates the file IF it does not already exist. The thing is even if internal artifactory probably dont want that token sitting here on the public repo so should perhaps be put in a configmap or secret that the image will read and use.

resources/common/start-custom.sh
```diff
@@ -137,18 +137,26 @@ if [ ! -d "$CS_DEFAULT_HOME/Machine" ]; then
...
+conda config --add channels $serviceaccountname:$serviceaccounttoken@https://artifactory.cloud.statcan.ca/artifactory/conda-forge-remote/
+conda config --add channels $serviceaccountname:$serviceaccounttoken@https://artifactory.cloud.statcan.ca/artifactory/conda-nvidia-remote/
+conda config --add channels $serviceaccountname:$serviceaccounttoken@https://artifactory.cloud.statcan.ca/artifactory/conda-pytorch-remote/
+
+pip config set global.index-url $serviceaccountname:$serviceaccounttoken@https://artifactory.cloud.statcan.ca/artifactory/pypi-remote/
+
+# if rprofile doesnt exist
+if [ ! -d "/opt/conda/lib/R/etc/Rprofile.site" ]; then
+  echo "Creating rprofile"
+  cat > test.txt<< EOF
```
Jose-Matsuda, line 153: Reminder change this to Rprofile.site oops my bad

Jose-Matsuda (approved):

Jose-Matsuda:
silly timeout error

PR #613, opened 2024-04-19
remove mc wrapper

bryanpaget:
I have removed the mc wrapper since it is no longer required.


bryanpaget:
I made a mistake, this should be two separate commits.

PR #615, opened 2024-04-23
remove mc-tenant-wrapper.sh and references to it

bryanpaget:
`mc-tenant-wrapper.sh` has been removed.


bryanpaget:
This builds, and I am able to test it on dev but I am not able to test it on The Zone.

tests/general/test_packages.py
```diff
@@ -96,6 +96,9 @@
     # import yaml, not import pyyaml
     "pyyaml",
     "graphviz",
+    # tables and pytables added because causing issues, TODO: understand issue and resolve
+    "tables",
+    "pytables"
```
mathis-marcotte, line 101: pytables is not a package

mathis-marcotte (changes requested):

.github/workflows/build_push.yaml
```diff
@@ -86,8 +86,8 @@ jobs:
     - name: Set ENV variables for a PR containing the auto-deploy tag
       if: github.event_name == 'pull_request' && contains( github.event.pull_request.labels.*.name, 'auto-deploy')
       run: | 
-        echo "REGISTRY=k8scc01covidacrdev.azurecr.io" >> "$GITHUB_ENV"
-        echo "IMAGE_VERSION=dev" >> "$GITHUB_ENV"
+        echo "REGISTRY=k8scc01covidacr.azurecr.io" >> "$GITHUB_ENV"
+        echo "IMAGE_VERSION=testing-mc-tenant-wrapper" >> "$GITHUB_ENV"
```
mathis-marcotte, line 90: you shouldnt change this image version. Just the registry change should be fine. Your image will be tagged with both "dev" and the branch name later in the workflow

mathis-marcotte (changes requested):

bryanpaget:
I tested this in The Zone, I created a simple file called data.csv with one row and three columns, I copied it to the bucket, deleted the original, copied the file from the bucket back to my home dir. Success.

```(base) jovyan@remove-mc-tenant-wrapper-0:~$ mc cp myBucketAlias/s3bucket/data.csv ./data.csv --insecure
...ud.statcan.ca/s3bucket/data.csv: 18 B / 18 B ┃▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓┃ 111 B/s 0s(base) jovyan@remove-mc-tenant-wrapper-0:~$ ```

tests/general/test_packages.py
bryanpaget: the package is called pytables and because of the way this test was designed it will try to import pytables, hence the above mapping between package names and import names. I could remove tables from that ignore list. That's why pyyaml is there. pyyaml also suffers a similar fate of not being the import name.

mathis-marcotte: That is true, I was just basic my thoughts on this https://pypi.org/project/tables/, but i guess pytables is the "english" name for it. Not much harm in keeping it in this list them, especially if this is a temporary change until we really fix the issue.

PR #616, opened 2024-04-23
Add font dev packages for r dependencies

bryanpaget:
These dev packages are required by some important R packages:

image

More info can be found here:
- https://jirab.statcan.ca/browse/BTIS-235


chuckbelisle (approved):
Trivy scan passed and all seems good to go. Once deployed, please confirm that the expected behaviour is what we get from the change.

bryanpaget:
Testing failed. I might need to install:

- libtiff-dev
- libfreetype-dev
- libfreetype6

PR #617, opened 2024-05-01
Bryan alternate sas base

bryanpaget:
This is a test to see if I can build the SAS image without being based on the SAS image (just copying the SAS files into the appropriate place).


bryanpaget:
This test was successful when run on my VM, Github is still having some issues. Closing PR.

PR #625, opened 2024-05-15
Add pkg config to conda

bryanpaget:
This was necessary to actually fix https://jirab.statcan.ca/browse/BTIS-235.

R seems to require that we have pkg-config installed by mamba/conda.

- https://stackoverflow.com/questions/68824450/error-configuration-failed-for-package-ragg

### DIFF
--- a/tests/general/test_packages.py
+++ b/tests/general/test_packages.py
@@ -97,7 +97,8 @@ EXCLUDED_PACKAGES = [
     # pytables has been added because it is causing issues, TODO: understand issue and resolve
     # pytables is the package but we import tables
     "tables",
-    "pytables"
+    "pytables",
+    "pkg-config"
 ]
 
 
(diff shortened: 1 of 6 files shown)


tests/general/test_packages.py
```diff
@@ -97,7 +97,8 @@
     # pytables has been added because it is causing issues, TODO: understand issue and resolve
     # pytables is the package but we import tables
     "tables",
-    "pytables"
+    "pytables",
+    "pkg-config"
```
mathis-marcotte, line 101: Is there a reason for adding this package to the list of excluded packages?

bryanpaget: It was tripping up the package test.

https://github.com/StatCan/aaw-kubeflow-containers/actions/runs/9100771771/job/25016384934#step:20:824

mathis-marcotte: Ok got it! That was my assumption.

PR #635, opened 2024-05-30
Rename and add env var

wg102:
Relates to https://jirab.statcan.ca/browse/BTIS-269


Souheil-Yazji (changes requested):
Approve after removal of "zone" from file name

Souheil-Yazji (approved):

PR #682, opened 2024-11-01
Update trivy

EveningStarlight:
Updates trivy to the latest version.
uses new `--db-repository` and `--java-db-repository` flags to point to point to multiple mirrors of the databases. 
This should mean many fewer `TOOMANYREQUESTS` errors.


mathis-marcotte (approved):
lgtm. Looks like all the build failures are unrelated.

PR #687, opened 2024-11-08
Add retry functionality to Trivy scan

EveningStarlight:
Changed trivy error code to 10. This means if vulnerabilities are found, it will error with exit code 10. 
This lets us retry any scan that had an error that was not finding vulnerabilities, since general errors are code 1. 

Will retry up to 5 times. 

The code is written like this as trivy doesn't actually write anything to an output variable on return, so there isn't any way to look for the `TOOMANYREQUESTS` error specifically. 

Closes https://github.com/StatCan/aaw/issues/1981

### DIFF
--- a/.github/workflows/build_push.yaml
+++ b/.github/workflows/build_push.yaml
@@ -187,7 +187,7 @@ jobs:
 
           trivy image \
             ${{ steps.build-image.outputs.full_image_name }} \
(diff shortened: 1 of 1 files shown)


EveningStarlight:
Retry works successfully now

image

Jose-Matsuda (approved):

PR #713, opened 2024-12-16
feat(stages)!: staged builds including final image tests

EveningStarlight:
PR used only to trigger CI, 
later PR will be the actual content to be merged. 

testing running of final tests, 
and pulling all existing images for speed since there's no diff compared with staged-builds-revised-stages


EveningStarlight:
testing finished

PR #720, opened 2025-01-03
feat: add ktutil and a helper script to our images

mathis-marcotte:
# Description

for https://jirab.statcan.ca/browse/BTIS-700 

# Tests / Quality Checks

Can be tested by running "ktutil-keytab" in a terminal in a running notebook

## Are there breaking changes?
Ask yourself the next question;
- [ ] Do we want to maintain the previous image from which we had to do breaking changes from?

If no, then carry on. If yes, there is a breaking change and we **want to maintain the previous image** do the following
- [ ] Create a new branch for the current version (ex v1) based off the current master/main branch
- [ ]  Increment the tag in the CI for pushes to master/main (v1 to v2)
- [ ] Change the CI that on pushes to the newly created "v1" branch (the name of the newly created branch we want to maintain is) it will push to the ACR. 
## Automated Testing/build and deployment
- [ ] Does the image pass CI successfully (build, pass vulnerability scan, and pass automated test suite)?
- [ ] If new features are added (new image, new binary, etc), have new automated tests been added to cover these?
- [ ] If new features are added that require in-cluster testing (e.g. a new feature that needs to interact with kubernetes), have you added the `auto-deploy` tag to the PR before pushing in order to build and push the image to ACR so you can test it in cluster as a custom image?

## JupyterLab extensions

- [ ] Are all extensions "enabled" (`jupyter labextension list` from inside the notebook)?

## VS Code tests

- [ ] Does VS Code open?
- [ ] Can you install extensions?

## Code review

- [ ] Have you added the `auto-deploy` tag to your PR before your most recent push to this repo?  This causes CI to build the image and push to our ACR, letting reviewers access the built image without having to create it themselves

### DIFF
--- a/docker-bits/4_CLI.Dockerfile
+++ b/docker-bits/4_CLI.Dockerfile
@@ -19,6 +19,8 @@ RUN apt-get update && \
       'zip' \
       'zsh' \
       'dos2unix' \
+      # installs necessary tool for kerberos authentication setup
+      'krb5-user' \
       # these are required by some r packages, adding these here so they get
       # installed into all images.
       'libfreetype6-dev' \
@@ -96,3 +98,5 @@ RUN wget -q "${GIT_CRED_MANAGER_URL}" -O ./gcm.deb \
   && echo "${GIT_CRED_MANAGER_SHA}  ./gcm.deb" | sha256sum -c - \
   && dpkg -i ./gcm.deb \
   && rm ./gcm.deb
+
+COPY ktutil-keytab.sh /home/$NB_USER/.local/bin/ktutil-keytab
--- a/docker-bits/4_CLI.Dockerfile
+++ b/docker-bits/4_CLI.Dockerfile
@@ -99,4 +99,6 @@ RUN wget -q "${GIT_CRED_MANAGER_URL}" -O ./gcm.deb \
   && dpkg -i ./gcm.deb \
   && rm ./gcm.deb
 
+# add script for kerberos keytab creation
 COPY ktutil-keytab.sh /usr/local/bin/ktutil-keytab
+RUN chmod +x /usr/local/bin/ktutil-keytab
--- a/docker-bits/4_CLI.Dockerfile
+++ b/docker-bits/4_CLI.Dockerfile
@@ -99,4 +99,4 @@ RUN wget -q "${GIT_CRED_MANAGER_URL}" -O ./gcm.deb \
   && dpkg -i ./gcm.deb \
   && rm ./gcm.deb
 
-COPY ktutil-keytab.sh /home/$NB_USER/.local/bin/ktutil-keytab
+COPY ktutil-keytab.sh /usr/local/bin/ktutil-keytab
--- a/output/docker-stacks-datascience-notebook/ktutil-keytab.sh
+++ b/output/docker-stacks-datascience-notebook/ktutil-keytab.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+# creates the kerberos directory if not exist
+mkdir -p ~/krb5
+cd ~/krb5
+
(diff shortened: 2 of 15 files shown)


output/docker-stacks-datascience-notebook/ktutil-keytab.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+# creates the kerberos directory if not exist
+mkdir -p ~/krb5
+cd ~/krb5
+
+# gets the user's username (legacy AD)
+read -p "Username(ex. marcoma):" user_name
+
+user_name="${user_name}@STATCAN.CA"
```
wg102, line 9: should it be @statcan.gc.ca ?

wg102 (commented):
Just one comment and if you could confirm the following:
This will trigger for everyone and every notebook?

output/docker-stacks-datascience-notebook/ktutil-keytab.sh
mathis-marcotte: I understand the concern but nope. This is correct. https://gitlab.k8s.cloud.statcan.ca/cloudnative/docker/kerberos-sidecar/-/tree/master?ref_type=heads#parameters

mathis-marcotte:
This script command will be available on every notebook type

docker-bits/4_CLI.Dockerfile
```diff
@@ -96,3 +98,7 @@ RUN wget -q "${GIT_CRED_MANAGER_URL}" -O ./gcm.deb \
   && echo "${GIT_CRED_MANAGER_SHA}  ./gcm.deb" | sha256sum -c - \
   && dpkg -i ./gcm.deb \
   && rm ./gcm.deb
+
+# add script for kerberos keytab creation
+COPY ktutil-keytab.sh /usr/local/bin/ktutil-keytab
+RUN chmod +x /usr/local/bin/ktutil-keytab
```
wg102, line 104: Should there be a prompt/instructions for the user to know to run this?

wg102 (commented):
Random not a specific place to write comments:
1. What if the user enters the wrong password? Does it update?
2. What if they use the keytab with a wrong password, what happens
3. is one keytab valid for all notebooks, (I think so) if so, does the file appear/is visible without the user needing to do the command. 
4. If not, does it mean only one is valid at a time
5. I don't think this is related to your ticket but any chance some of this can be cleaned up : 
image
6. What if the notebook is not persistent?

docker-bits/4_CLI.Dockerfile
mathis-marcotte: There is a ticket to define those instructions in the docs to let users know what to do 
https://jirab.statcan.ca/browse/BTIS-686

mathis-marcotte:
1. What if the user enters the wrong password? Does it update?
Yes it does. This script doesn't do any actual authentication. I don't know if there is a way to even do that. This basically just creates their credentials that will be used when it is time to authenticate

2. What if they use the keytab with a wrong password, what happens
Same question as 1? So same answer

3. is one keytab valid for all notebooks, (I think so) if so, does the file appear/is visible without the user needing to do the command.
Yes. Once the command is run, it creates the file under /~/krb5, so yes it can be visible. This new command doesn't display the keytab. (Users shouldn't need to see this file anyways)

4. If not, does it mean only one is valid at a time
Irrelevant based on previous question's answer. But one thing to note, the keytab becomes invalid if a user's password changes. So when their netA password changes, they will need to run this again.

5. I don't think this is related to your ticket but any chance some of this can be cleaned up
Possibly.

6. What if the notebook is not persistent?
Not sure what this means. The command creates a k8s secret with the keytab value. We have big problems if somehow, k8s secrets aren't persisted.

PR #727, opened 2025-01-16
Staged images 1 GitHub workflows

EveningStarlight:
Part one of the migration to staged builds. 

This part includes all changes to the github workflows, and some changes to the makefile to support the new functionality. 

A diagram of the relationships is shown below
Workflows


.github/workflows/docker-pull-test.yaml
```diff
@@ -0,0 +1,111 @@
...
+      - name: Set up venv for Test Suite
+        run: |
+          python -m pip install --upgrade pip
+          make install-python-dev-venv
+
+      - name: Test image
+        run: make test/${{ inputs.image }}
+
+      # Free up space from build process (containerscan action will run out of space if we don't)
+      - name: cleanup runner``
```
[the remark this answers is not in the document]
EveningStarlight: nope, will remove

.github/workflows/docker-build-upload.yaml
```diff
@@ -0,0 +1,116 @@
...
+          - 5000:5000
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Run Hadolint
+        run: |
+          sudo curl -L https://github.com/hadolint/hadolint/releases/download/v${{ env.HADOLINT_VERSION }}/hadolint-Linux-x86_64 --output hadolint
+          sudo chmod +x hadolint
+          ./hadolint images/${{ inputs.directory }}/Dockerfile --no-fail
```
Souheil-Yazji, line 61: I'm not opposed to failing on lint

.github/workflows/docker-build-upload.yaml
```diff
@@ -0,0 +1,116 @@
...
+
+      - name: Set BASE_IMAGE variable
+        run: |
+          if [ "${{ inputs.base-image }}" == "" ]; then
+            echo "BASE_IMAGE=${{ steps.pull-parent.outputs.image_name }}" >> $GITHUB_ENV
+          else
+            echo "BASE_IMAGE=${{ inputs.base-image }}" >> $GITHUB_ENV
+          fi
+
+      - name: Set FROM and as in Docerfile
```
Souheil-Yazji, line 92: typo Dockerfile

.github/workflows/docker-nightly.yaml
```diff
@@ -0,0 +1,85 @@
+name: Nightly test
+on:
+  schedule:
+    # Execute at 2am EST every day
+    - cron:  '0 7 * * *'
```
Souheil-Yazji, line 5: 7 utc is accurate!

.github/workflows/docker-pull-test.yaml
```diff
@@ -0,0 +1,111 @@
...
+        run: make test/${{ inputs.image }}
+
+      # Free up space from build process (containerscan action will run out of space if we don't)
+      - name: cleanup runner
+        run: ./.github/scripts/cleanup_runner.sh
+
+      # Scan image for vulnerabilities
+      - name: Aqua Security Trivy image scan
+        run: |
+          curl -sfL https://raw.githubusercontent.com/aquasecurity/trivy/main/contrib/install.sh | sh -s -- -b /usr/local/bin ${{ env.TRIVY_VERSION }}
```
Souheil-Yazji, line 84: would love to see a checksum since we are pulling this install script from the internet. Otherwise I think it's fine.

Souheil-Yazji: I don't think there's one offered upstream :( we can pull the script, look at it (confirm it's fine in quality), compute the sha, then use that as our reference. Then compute the SHA on retrieval from the internet and compare.

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,259 @@
...
+      - name: Set up environment
+        run: echo "Environment has been set up."
+
+  base:
+    needs: [vars]
+    uses: ./.github/workflows/docker-steps.yaml
+    with:
+      image: "base"
+      directory: "base"
+      base-image: "quay.io/jupyter/datascience-notebook:2024-06-17"
```
Souheil-Yazji, line 50: I assume if we want to update the base image, we have to do it here as well as the Base dockerfile

Souheil-Yazji (approved):
Approved with Comments. 🚀

Souheil-Yazji:
@mathis-marcotte  to provide a review before merge.

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,259 @@
...
+      # images for hotfixes, so always pin tag and digest to prevent unexpected upstream changes
+      base-image: "rocker/geospatial:4.2.1@sha256:5caca36b8962233f8636540b7c349d3f493f09e864b6e278cb46946ccf60d4d2"
+      registry-name: "${{ needs.vars.outputs.REGISTRY_NAME }}"
+      buildkit: 0
+      branch-name: "${{ needs.vars.outputs.branch-name }}"
+    secrets:
+      REGISTRY_USERNAME: ${{ secrets.REGISTRY_USERNAME }}
+      REGISTRY_PASSWORD: ${{ secrets.REGISTRY_PASSWORD }}
+
+  # Test the Images
```
mathis-marcotte, line 199: Any way that this section and the "docker-nightly.yaml" workflow can share code? 
They both seem to just run "docker-pull-test" on the list of our images. Although i do see that the "needs" section here is more than just "vars". 
This is making me thing that in this current version of the workflow file, these tests can be run as soon as the build step is completed for that particular image. In a scenario where code would be shared between this and the nightly build workflow, that might change the behavior and the tests would need to be ran after all the builds are complete, which sounds like would be a longer workflow action in the end.
Anyways, not saying this absolutely needs to change, just wanted to get you thinking if we could limit code duplication between files.

mathis-marcotte (changes requested):
I think that at the very least, the docker-nightly comment I left should be addressed. Outside of that, LGTM, i just left some thoughts/ideas. Up to you on if they are good thoughts or not.

.github/workflows/docker.yaml
EveningStarlight: We would actually only do it here. 
It's set this way because the `From *** as ***` cannot take any variables,
so the code is set to inject the line dynamically from the variables found in the `docker.yaml` file.

.github/workflows/docker.yaml
EveningStarlight: I can't think of a way to solve the trade off between reducing code duplication and running the tests as the image completes. 

The tests all running at the end shouldn't effect the overall workflow time much, as they are roughly all the same time as each other. But will increase the time a developer has to wait for the results of the shorter build images to be tested. 
Maybe that would be worth the trade off though?

.github/workflows/docker-build-upload.yaml
```diff
@@ -0,0 +1,116 @@
...
+        run: |
+          sed -i '1i FROM ${{ env.BASE_IMAGE}} as ${{ inputs.image }}' ./images/${{ inputs.directory }}/Dockerfile
```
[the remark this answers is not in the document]
EveningStarlight: `1i` means that sed is used to insert the line into the file
https://stackoverflow.com/questions/28116144/how-to-insert-a-line-at-the-beginning-of-a-file-with-sed

None of the Dockerfiles have a FROM in them at all, they are all injected into the file using this sed command.

PR #728, opened 2025-01-16
Staged images 2 image folder

EveningStarlight:
Part two of the migration to staged builds.

Image folder is comprised of a subdirectory for each stage used by the build. 
Each subdirectory contains the necessary Dockerfile, resources, and scripts to build the stage. 

Dockerfiles are composed of the dockerbits that were previously used. 
Rest of files are copy past from other locations.
Only new addition is a copy line to the base that adds a requirements.txt to `usr/local/bin` that will install components required for using pytorch


Jose-Matsuda (commented):
Only comment I have is what Souheil suggested in having utilities hosted in a pullable way so the repo doesn't have as many files but that's not a blocker / for the beginning it might be better to have everything super obvious.

Souheil-Yazji (approved):
LGTM

Souheil-Yazji:
> Only comment I have is what Souheil suggested in having utilities hosted in a pullable way so the repo doesn't have as many files but that's not a blocker / for the beginning it might be better to have everything super obvious.

For this, we'll just do it in another issue. Keeps scope clear

Issue #732, opened 2025-01-22
Resolve issue mentioned in Staged Build PRs


EveningStarlight:
Work linked to:
https://github.com/StatCan/aaw-kubeflow-containers/issues/705
https://github.com/StatCan/aaw-kubeflow-containers/issues/707
https://github.com/StatCan/aaw-kubeflow-containers/issues/709

Comments made:

Workflows, https://github.com/StatCan/aaw-kubeflow-containers/pull/727

- [x] fail hadolint?
- [x] docker-build-upload.yaml fix typo - Docerfile
- [x] Set a SHA for trivy 
- [x] docker-nightly.yaml should be a matrix
- [ ] ~~reuse code from nightly for after build ? tradeoff~~

Documentation, https://github.com/StatCan/aaw-kubeflow-containers/pull/729

- [x] update the staged build documentation image. make base match cmd and platform

Other

- [ ] ~~Fail trivy scans again once builds prove successful~~ moved to https://github.com/StatCan/aaw-kubeflow-containers/issues/734
- [x] clean all old branches and fork

EveningStarlight:
Closed by https://github.com/StatCan/aaw-kubeflow-containers/pull/736

PR #735, opened 2025-01-22
Enable Hadolint to fail builds

EveningStarlight:
investigates removing `--no-fail` from Hadolint

Part of https://github.com/StatCan/aaw-kubeflow-containers/issues/732

Hadolint checks had to be moved below the `FROM *** as ***` injection, or would else error out on the missing `FROM`

Fixes made for

- fixes https://github.com/hadolint/hadolint/wiki/DL3003 (Use absolute paths, or use WORKDIR to switch to a directory)
- fixes https://github.com/hadolint/hadolint/wiki/DL3004 (Do not use sudo)
- fixes https://github.com/hadolint/hadolint/wiki/DL3005 (Do not use apt-get dist-upgrade)
- fixes https://github.com/hadolint/hadolint/wiki/DL3027 (Do not use apt, it is an end-user tool, use apt-get instead)
- fixes https://github.com/hadolint/hadolint/wiki/DL3042 (Avoid cache directory with pip install --no-cache-dir)
- fixes https://github.com/koalaman/shellcheck/wiki/SC2086 (Double quote to prevent globbing and word splitting)

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -1,3 +1,4 @@
+FROM quay.io/jupyter/datascience-notebook:2024-06-17 as base
 USER root
 
 ENV PATH="/home/jovyan/.local/bin/:${PATH}"
@@ -82,7 +83,7 @@ RUN \
     && curl -LO "${KUBECTL_URL}.sha256" \
     && echo "$(cat kubectl.sha256) kubectl" | sha256sum -c - \
     && chmod +x ./kubectl \
-    && sudo mv ./kubectl /usr/local/bin/kubectl \
+    && mv ./kubectl /usr/local/bin/kubectl \
   && \
     # AzureCLI - installation script from Azure
     curl -sLO "${AZCLI_URL}" \
@@ -100,7 +101,7 @@ RUN \
     && echo "${TRINO_SHA}" /tmp/trino-original | sha256sum -c \
     && echo "trinocli: ok" \
     && chmod +x /tmp/trino-original \
-    && sudo mv /tmp/trino-original /usr/local/bin/trino-original \
+    && mv /tmp/trino-original /usr/local/bin/trino-original \
   && \
     # argo cli
     curl -sLO  "${ARGO_CLI_URL}" \
@@ -108,7 +109,7 @@ RUN \
     && grep argo-linux-amd64.gz argo-workflows-cli-checksums.txt | sha256sum -c - \
     && gunzip argo-linux-amd64.gz \
     && chmod +x argo-linux-amd64 \
-    && sudo mv ./argo-linux-amd64 /usr/local/bin/argo \
+    && mv ./argo-linux-amd64 /usr/local/bin/argo \
     && argo version \
   && \
     # quarto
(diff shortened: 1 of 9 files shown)


EveningStarlight:
Current errors to resolve

```
 images/base/Dockerfile:73 DL3004 error: Do not use sudo as it leads to unpredictable behavior. Use a tool like gosu to enforce root

images/remote-desktop/Dockerfile:129 DL3004 error: Do not use sudo as it leads to unpredictable behavior. Use a tool like gosu to enforce root
images/remote-desktop/Dockerfile:421 DL3004 error: Do not use sudo as it leads to unpredictable behavior. Use a tool like gosu to enforce root
images/remote-desktop/Dockerfile:444 DL3004 error: Do not use sudo as it leads to unpredictable behavior. Use a tool like gosu to enforce root

images/rstudio/Dockerfile:5 DL3004 error: Do not use sudo as it leads to unpredictable behavior. Use a tool like gosu to enforce root
images/rstudio/Dockerfile:5 DL3005 error: Do not use apt-get dist-upgrade

images/tensorflow/Dockerfile:19 DL3004 error: Do not use sudo as it leads to unpredictable behavior. Use a tool like gosu to enforce root
```

images/base/force-build.txt
mathis-marcotte, line 1: Is this file necessary? I dont see it used anywhere

images/remote-desktop/force-build.txt
mathis-marcotte, line 1: again, empty file here

images/base/Dockerfile
```diff
@@ -97,36 +97,36 @@ RUN \
...
-    && chmod +x quarto-${QUARTO_VERSION} \
-    && sudo rm -f /usr/local/bin/quarto \
-    && sudo mv ./quarto-${QUARTO_VERSION} /usr/local/bin/quarto 
+    curl -sLO  "${QUARTO_URL}" \
+    && curl -LO "${QUARTO_CHECKSUM_URL}" \
+    && grep "quarto-${QUARTO_VERSION}-linux-amd64.tar.gz" "quarto-${QUARTO_VERSION}-checksums.txt" | sha256sum -c - \
+    && tar -xf "quarto-${QUARTO_VERSION}-linux-amd64.tar.gz" \
+    && chmod +x "quarto-${QUARTO_VERSION}" \
+    && rm -f /usr/local/bin/quarto \
+    && mv "./quarto-${QUARTO_VERSION}" /usr/local/bin/quarto
```
mathis-marcotte, line 121: Have you tested that removing sudo does not impact the installation of packages? Not talking about just quarto, just talking generally since it seems like sudo was removed in a couple places (I'm assuming this was hadolint that said to do this)

mathis-marcotte (changes requested):
A couple empty files with no reference were commited

images/base/force-build.txt
EveningStarlight: It was necessary in order to create a diff so that the files would build and be tested by hadolint. 
Will be removed before the branch gets pulled in.

images/base/Dockerfile
EveningStarlight: On review I saw that one sudo that I removed was an install, so I put that back. But I did check the logs and sudo is already installed and up to date. So I'm not sure why it's there. But that's a separate issue. 

The remaining `sudo`s were used to `mv` or `rm` files/folder. If these are unable to be completed they will fail the build. 
Since the build completes and passes the package tests after `sudo` was removed, I concluded that they are still installed correctly. 
But I didn't look interactively to double check.

mathis-marcotte (approved):
lgtm

PR #744, opened 2025-01-28
roll back OpenM++ version

EveningStarlight:
Fix issue in https://github.com/StatCan/aaw/issues/2016

The newer version of OpenM++ need `glibc v2.38`, but on our current version we have `glibc v2.35`
This version is set by the Ubuntu version, which is set by our base image. 

We can upgrade OpenM++ once https://github.com/StatCan/aaw/issues/2014 is finished.


Souheil-Yazji (approved):
LGTM 🚀

PR #754, opened 2025-03-10
fix: tags

EveningStarlight:
Some places are using branch name as tag. This is unreliable for master as it is the same as The Zone. 
We must use `v1` when referencing the master branch for AAW. (and `v2` for The Zone. )

changed the variable `branch-name` to `tag` for testing, where it better describes the process.


jacek-dudek (approved):
Looks good to me.

PR #764, opened 2026-04-20
update: Trivy version

EveningStarlight:
Updating the Trivy image version. 
Also helps create a newer image, so that the Zone team can clean up old images.


EveningStarlight:
* Updated Trivy version
* pinned specific julia package for base image
* downloads and builds the pspp package

